PR #5, opened 2024-04-25
Sort commatrix by node role, protocol and port

sabinaaledort:
Sort commatrix by node role, protocol and port and fix ss filter causing missing UDP ports


types/types.go
```diff
@@ -15,25 +17,23 @@ type ComMatrix struct {
...
-	NodeRole  string `json:"nodeRole"`
-	Optional  bool   `json:"optional"`
+	Direction string `json:"Direction"`
+	Protocol  string `json:"Protocol"`
+	Port      int    `json:"Port,string"`
+	Namespace string `json:"Namespace"`
+	Service   string `json:"Service"`
+	Pod       string `json:"Pod"`
+	Container string `json:"Container"`
+	NodeRole  string `json:"Node Role"`
```
oribon, line 27: I think the space here is wrong (in the `json:`), also is there a reason for the uppercase change in the rest?

types/types.go
```diff
@@ -78,14 +78,14 @@ func (m *ComMatrix) String() string {
 }
 
 func (cd ComDetails) String() string {
-	return fmt.Sprintf("%s,%s,%s,%s,%s,%s,%s,%s,%v", cd.Direction, cd.Protocol, cd.Port, cd.Namespace, cd.Service, cd.Pod, cd.Container, cd.NodeRole, cd.Optional)
+	return fmt.Sprintf("%s,%s,%d,%s,%s,%s,%s,%s,%v", cd.Direction, cd.Protocol, cd.Port, cd.Namespace, cd.Service, cd.Pod, cd.Container, cd.NodeRole, cd.Optional)
 }
 
 func RemoveDups(outPuts []ComDetails) []ComDetails {
```
oribon, line 84: nit: it seems this and the `CmpComDetails` are called together, wdyt about just combining them into one function that remove dups and sorts?

cmd/main.go
```diff
@@ -171,26 +176,35 @@ func main() {
...
+
+	err = os.WriteFile(filepath.Join(destDir, "matrix-diff-ss"),
+		[]byte(diff),
+		0644)
+	if err != nil {
+		panic(err)
+	}
+}
+
+func diff(mat types.ComMatrix, ssMat types.ComMatrix) string {
```
oribon, line 189: the name can be changed to something clearer, also is ssMat still correct here? can this just be mat1,mat2?

oribon (commented):
lgtm

PR #7, opened 2024-05-06
Allow setting custom-entries-file in formats

liornoy:
This commit adds the the flag customEntriesFormat that sets the custom entries file format (json/yaml/csv).

We also include 3 example files to help the user to create his own custom file, and they'll be used in the integration test.


commatrix/commatrix.go
```diff
@@ -34,7 +37,7 @@ const (
 // Custom entries from a JSON file can be added to the matrix by setting `customEntriesPath`.
 // Returns a pointer to ComMatrix and error. Entries include traffic direction, protocol,
 // port number, namespace, service name, pod, container, node role, and flow optionality for OpenShift.
-func New(kubeconfigPath string, customEntriesPath string, e Env, d Deployment) (*types.ComMatrix, error) {
+func New(kubeconfigPath string, customEntriesPath string, customEntriesFormat string, e Env, d Deployment) (*types.ComMatrix, error) {
```
sabinaaledort, line 40: customEntriesFormat should be an enum

commatrix/commatrix.go
```diff
@@ -61,7 +64,10 @@ func New(kubeconfigPath string, customEntriesPath string, e Env, d Deployment) (
 	res = append(res, staticEntries...)
 
 	if customEntriesPath != "" {
-		customComDetails, err := addFromFile(customEntriesPath)
+		if customEntriesFormat == "" {
```
sabinaaledort, line 67: Should be validated in main

types/types.go
```diff
@@ -51,7 +51,7 @@ func ToCSV(m ComMatrix) ([]byte, error) {
 }
 
 func ToJSON(m ComMatrix) ([]byte, error) {
-	out, err := json.Marshal(m.Matrix)
+	out, err := json.MarshalIndent(m.Matrix, "", "    ")
```
sabinaaledort, line 54: Why this change?

commatrix/commatrix.go
liornoy: done, the flag still accepts string, but internally it's validated as an enum

commatrix/commatrix.go
liornoy: done

PR #10, opened 2024-06-19
Add filter condition for the diff matrix

liornoy:
This commit adds filter condition to the diff matrix
to make sure that the diff is not shwoing irrelevant
entries such as rpc.statd or dynamic node port ports
of the `ovnkube` service.


liornoy:
closing this one, because we will apply this filter on the origin coverage test side.

PR #14, opened 2024-07-01
Generate the communication matrix in nftables format

aabughosh:
in this PR there is few things 
main thing is 1.adding option of  FORMAT=nft to create the communication  matrix in output file as .nft file
user can do nft apply -f to that specific  file
2.few arrangements  to the code separate  into functions and files


sabinaaledort:
I think for a user, generating the matrix in nftables format and applying them should be two option in the Makefile and not done by default

cmd/main.go
```diff
@@ -185,6 +185,28 @@ func main() {
 	if err != nil {
 		panic(err)
 	}
+
+	err = commatrix.ApplyFireWallRules(cs, mat, "master")
+	if err != nil {
+		panic(err)
+	}
+
+	err = commatrix.ApplyFireWallRules(cs, mat, "worker")
```
sabinaaledort, line 194: If SNO apply on master only

Makefile
```diff
@@ -32,5 +32,8 @@ generate: oc build
 	mkdir -p $(DEST_DIR)/communication-matrix
 	./$(EXECUTABLE) -format=$(FORMAT) -env=$(CLUSTER_ENV) -destDir=$(DEST_DIR)/communication-matrix -deployment=$(DEPLOYMENT) -customEntriesPath=$(CUSTOM_ENTRIES_PATH) -customEntriesFormat=$(CUSTOM_ENTRIES_FORMAT)
 
+apply-firewall: build
+	./$(EXECUTABLE) -applyFirewall -format=$(FORMAT) -env=$(CLUSTER_ENV) -destDir=$(DEST_DIR)/communication-matrix -deployment=$(DEPLOYMENT)
```
sabinaaledort, line 36: format and destDir are not used

sabinaaledort: sorry destDir is used but I'm wondering if apply command should just be apply and allow to generate the mat in nftables format

aabughosh: thats whats doing its applying the firewall by that mat and generate the nftformat

commatrix/generate.go
```diff
@@ -0,0 +1,175 @@
...
+
+	clientutil "github.com/openshift-kni/commatrix/client"
+	"github.com/openshift-kni/commatrix/consts"
+	"github.com/openshift-kni/commatrix/debug"
+	"github.com/openshift-kni/commatrix/ss"
+	"github.com/openshift-kni/commatrix/types"
+)
+
+// generate relevant matrix and diff files
+func GenerateCommatrix(kubeconfig, customEntriesPath, customEntriesFormat, format string, env Env, deployment Deployment, printFn func(m types.ComMatrix) ([]byte, error), destDir string) {
```
oribon, line 22: nit: this can be renamed to just `Generate` (the "commatrix" is inferred from the package name)

oribon, line 22: this should not receive a printFn (this is a pure implementation detail, the caller does not need to know that it exists) since the "format" is enough.
Also, it seems that the function does a lot more than just generating a matrix (generates, writes to file, diffs, more writes, etc.), I'd say it is worth splitting these functionalities to different functions and letting the caller decide what it wants (i.e this function should only return a matrix).

commatrix/generate.go
```diff
@@ -0,0 +1,175 @@
...
+	mat, err := New(kubeconfig, customEntriesPath, customEntriesFormat, env, deployment)
+	if err != nil {
+		panic(fmt.Sprintf("failed to create the communication matrix: %s", err))
```
oribon, line 25: We should not panic but rather return the err and let the caller decide what to do with it (same for the other panics)

types/types.go
```diff
@@ -160,3 +161,48 @@ func (m ComMatrix) Contains(cd ComDetails) bool {
...
+
+	tcpPortsStr := strings.Join(tcpPorts, ", ")
+	udpPortsStr := strings.Join(udpPorts, ", ")
+
+	result := fmt.Sprintf(`#!/usr/sbin/nft -f
+
+	table ip filter {
+		chain FIREWALL {
+			# Allow loopback traffic
+			type filter hook input priority 0; policy accept;
```
cgoncalves, line 184: Have you tried applying the generated nft output in your nftables?
It won't work, invalid syntax. Remove this line.

Here is a simple example:

```
table ip filter {
        chain FIREWALL {
                type filter hook input priority 0; policy accept;
        }
        chain INPUT {
                type filter hook input priority 0; policy accept;
                jump FIREWALL
        }
}
```


```
communication-matrix-master.nft:8:8-15: Error: Could not process rule: Operation not supported
		jump FIREWALL
		     ^^^^^^^^
```

commatrix/commatrix_test.go
```diff
@@ -0,0 +1,63 @@
...
+			if tt.expectedErr {
+				assert.Error(t, err)
+			} else {
+				assert.NoError(t, err)
+				assert.Equal(t, fmt.Sprintf("%T", fn), tt.expectedFnType)
+			}
+		})
+	}
+}
+func TestCreateOutputFiles(t *testing.T) {
```
sabinaaledort, line 37: nit: missing new line

commatrix/generate.go
```diff
@@ -0,0 +1,226 @@
...
+	case "yaml":
+		return types.ToYAML, nil
+	case "nft":
+		return types.ToNFTables, nil
+	default:
+		return nil, fmt.Errorf("invalid format: %s. Please specify json, csv, yaml, or nft", format)
+	}
+}
+
+func createOutputFiles(destDir string) (*os.File, *os.File, error) {
```
sabinaaledort, line 87: I think the name is too general, this opens raw ss files, I don't know if it really needs a function

commatrix/generate.go
```diff
@@ -0,0 +1,226 @@
...
+	udpFile, err := os.OpenFile(path.Join(destDir, "raw-ss-udp"), os.O_TRUNC|os.O_CREATE|os.O_WRONLY, 0644)
+	if err != nil {
+		tcpFile.Close()
+		return nil, nil, err
+	}
+
+	return tcpFile, udpFile, nil
+}
+
+func getSSmatrix(cs *clientutil.ClientSet, nodesList *v1.NodeList, tcpFile, udpFile *os.File) (*types.ComMatrix, error) {
```
sabinaaledort, line 102: this more create or generate

cmd/main.go
```diff
@@ -80,135 +58,13 @@ func main() {
...
-	defer tcpFile.Close()
-
-	udpFile, err := os.OpenFile(path.Join(destDir, "raw-ss-udp"), os.O_TRUNC|os.O_CREATE|os.O_WRONLY, 0644)
-	if err != nil {
-		panic(err)
-	}
-	defer udpFile.Close()
-
-	nodesList, err := cs.CoreV1Interface.Nodes().List(context.TODO(), metav1.ListOptions{})
+	ssComMat, endSliceMat, err := commatrix.GenerateMatrix(kubeconfig, customEntriesPath, customEntriesFormat, format, env, deployment, destDir)
```
sabinaaledort, line 61: GenerateMatrix  returns first the communication matrix only then the ss matrix:
`return mat, ssComMat, err`

commatrix/generate.go
```diff
@@ -0,0 +1,226 @@
...
+}
+
+// WriteMatsToFiles write all data to files(ss and endpoint slice matrix and the diff matrix).
+func WriteMatsToFiles(mat, ssComMat *types.ComMatrix, format string, env Env, deployment Deployment, destDir string) error {
+	printFn, err := getPrintFunction(format)
+	if err != nil {
+		return err
+	}
+
+	err = writeMatrixToFileByType(*mat, "communication-matrix", format, deployment, printFn, destDir)
```
sabinaaledort, line 59: I don't think both writeMatrixToFileByType and getPrintFunction are needed, you can just print the mat where you check the type

aabughosh: lets keep it as is

cmd/main.go
```diff
@@ -79,136 +56,34 @@ func main() {
 	if customEntriesPath != "" && customEntriesFormat == "" {
 		panic("error, variable customEntriesFormat is not set")
 	}
-
+	// 1.generate the endpointslice matrix
```
sabinaaledort, line 59: I don't think you have to number the steps and the comments is also not needed here since its clear enough that commatrix.New creating a commantrix

cmd/main.go
```diff
@@ -79,136 +56,34 @@ func main() {
...
-
-	comMatrixFileName := filepath.Join(destDir, fmt.Sprintf("communication-matrix.%s", format))
-	err = os.WriteFile(comMatrixFileName, res, 0644)
-	if err != nil {
-		panic(err)
+		panic(fmt.Errorf("failed to create the communication matrix: %s", err))
 	}
-
-	cs, err := clientutil.New(kubeconfig)
+	// 2.write the endpoint slice matrix to file
```
sabinaaledort, line 64: Same

cmd/main.go
```diff
@@ -79,136 +56,34 @@ func main() {
...
+	err = commatrix.WriteMatrixToFileByType(*mat, "communication-matrix", format, deployment, destDir)
 	if err != nil {
-		panic(err)
+		panic(fmt.Sprintf("Error while generating matrix :%v", err))
```
sabinaaledort, line 67: Need to update the error message

cmd/main.go
```diff
@@ -79,136 +56,34 @@ func main() {
...
 	}
-
-	tcpFile, err := os.OpenFile(path.Join(destDir, "raw-ss-tcp"), os.O_TRUNC|os.O_CREATE|os.O_WRONLY, 0644)
+	// 3.generate the ss matrix and ss raws
+	ssMat, ssOutTCP, ssOutUDP, err := commatrix.GenerateSS(kubeconfig, customEntriesPath, customEntriesFormat, format, env, deployment, destDir)
 	if err != nil {
-		panic(err)
+		panic(fmt.Sprintf("Error while generating matrix :%v", err))
```
sabinaaledort, line 72: Same

cmd/main.go
```diff
@@ -79,136 +56,34 @@ func main() {
...
 	}
-	defer tcpFile.Close()
-
-	udpFile, err := os.OpenFile(path.Join(destDir, "raw-ss-udp"), os.O_TRUNC|os.O_CREATE|os.O_WRONLY, 0644)
+	// 4.write ss files
+	err = commatrix.WriteSSRawFiles(destDir, ssOutTCP, ssOutUDP)
 	if err != nil {
-		panic(err)
+		panic(fmt.Sprintf("Error while generating matrix :%v", err))
```
sabinaaledort, line 77: Same

cmd/main.go
```diff
@@ -79,136 +56,34 @@ func main() {
...
 	}
-	defer udpFile.Close()
-
-	nodesList, err := cs.CoreV1Interface.Nodes().List(context.TODO(), metav1.ListOptions{})
+	// 5.write the ss matrix to file
+	err = commatrix.WriteMatrixToFileByType(*ssMat, "ss-generated-matrix", format, deployment, destDir)
 	if err != nil {
-		panic(err)
+		panic(fmt.Sprintf("Error while generating matrix :%v", err))
```
sabinaaledort, line 82: Same

commatrix/generate.go
```diff
@@ -0,0 +1,236 @@
...
+	if err != nil {
+		return nil, nil, nil, err
+	}
+
+	nodesList, err := cs.CoreV1Interface.Nodes().List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		return nil, nil, nil, err
+	}
+
+	ssComMat, ssOutTCP, ssOutUDP, err := generateSS(cs, nodesList)
```
sabinaaledort, line 33: Why did you separate to generateSS func that  is used just once?

commatrix/generate.go
```diff
@@ -0,0 +1,236 @@
...
+	if err != nil {
+		return nil, nil, nil, err
+	}
+	return ssComMat, ssOutTCP, ssOutUDP, err
+}
+
+// WriteMatsToFiles write all data to files(ss and endpoint slice matrix and the diff matrix).
+func WriteMatsToFiles(mat, ssComMat *types.ComMatrix, format string, env Env, deployment Deployment, destDir string) error {
```
sabinaaledort, line 41: Not used anymore

commatrix/generate.go
```diff
@@ -0,0 +1,236 @@
...
+	}
+	_, err = udpFile.Write([]byte(string(ssOutUDP)))
+	if err != nil {
+		return fmt.Errorf("failed writing to file: %s", err)
+	}
+
+	return nil
+}
+
+func WriteMatrixDiff(mat types.ComMatrix, ssComMat types.ComMatrix, fileName, destDir string) error {
```
sabinaaledort, line 158: Can just be a single function generateMatrixDiff

commatrix/generate.go
aabughosh: yah but the function is so much long this is why i seperate

commatrix/commatrix_test.go
```diff
@@ -0,0 +1,64 @@
...
+				assert.Error(t, err)
+			} else {
+				assert.NoError(t, err)
+				assert.Equal(t, fmt.Sprintf("%T", fn), tt.expectedFnType)
+			}
+		})
+	}
+}
+
+func TestCreateOutputFiles(t *testing.T) {
```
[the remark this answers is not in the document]
sabinaaledort: I don't think it's validating any functionally of the matrix module, only checking that linux can open files...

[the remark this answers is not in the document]
aabughosh: yes i will remove

commatrix/generate.go
sabinaaledort: I don't think 5 more lines are a reason to open another func with the same name, if you want to make the functions shorter you can split parts in it to functions for example the loop that's getting the ss from the node

ss/ss.go
```diff
@@ -24,23 +23,14 @@ const (
 	duration              = time.Second * 5
 )
 
-func CreateComDetailsFromNode(debugPod *debug.DebugPod, node *corev1.Node, tcpFile, udpFile *os.File) ([]types.ComDetails, error) {
-	ssOutTCP, err := debugPod.ExecWithRetry("ss -anpltH", interval, duration)
+func CreateComDetailsFromNode(debugPod *debug.DebugPod, node *corev1.Node) (res []types.ComDetails, ssOutTCP, ssOutUDP []byte, err error) {
```
sabinaaledort, line 26: The func called CreateComDetailsFromNode but returning ss raw outputs?

types/types.go
```diff
@@ -160,3 +161,47 @@ func (m ComMatrix) Contains(cd ComDetails) bool {
...
+		}
+	}
+
+	tcpPortsStr := strings.Join(tcpPorts, ", ")
+	udpPortsStr := strings.Join(udpPorts, ", ")
+
+	result := fmt.Sprintf(`#!/usr/sbin/nft -f
+
+	table ip filter {
+		chain FIREWALL {
```
sabinaaledort, line 182: let's call this chain OPENSHIFT instead of FIREWALL since these are ocp ports

commatrix/generate.go
aabughosh: i did on function its better

ss/ss.go
aabughosh: yes this is from lior i think also name is not good

types/types.go
aabughosh: ok

types/types.go
```diff
@@ -160,3 +161,47 @@ func (m ComMatrix) Contains(cd ComDetails) bool {
...
+			# Allow specific TCP and UDP ports
+			tcp dport  { %s } accept
+			udp dport { %s } accept
+
+			# Logging and default drop
+			log prefix "firewall " drop
+		}
+	
+		chain INPUT {
+			type filter hook input priority 0; policy accept;
```
cgoncalves, line 201: OpenShift/OVN-Kubernetes will also create a filter type hooked on input with default priority 0/zero.
Having that chain and this one hooked at input and with the same priority is NOT recommended.

From nftables documentation:

> It's possible to give two base chains the same priority, but there is no guaranteed evaluation order of base chains with identical priority that are attached to the same hook location

https://wiki.nftables.org/wiki-nftables/index.php/Configuring_chains

types/types.go
```diff
@@ -160,3 +161,47 @@ func (m ComMatrix) Contains(cd ComDetails) bool {
...
+			udpPorts = append(udpPorts, fmt.Sprint(line.Port))
+		}
+	}
+
+	tcpPortsStr := strings.Join(tcpPorts, ", ")
+	udpPortsStr := strings.Join(udpPorts, ", ")
+
+	result := fmt.Sprintf(`#!/usr/sbin/nft -f
+
+	table ip filter {
```
cgoncalves, line 181: The same table name is created by the cluster already. It is highly advisable to use a different name.

types/types.go
```diff
@@ -160,3 +161,47 @@ func (m ComMatrix) Contains(cd ComDetails) bool {
...
+			tcp dport  { %s } accept
+			udp dport { %s } accept
+
+			# Logging and default drop
+			log prefix "firewall " drop
+		}
+	
+		chain INPUT {
+			type filter hook input priority 0; policy accept;
+			jump FIREWALL
```
cgoncalves, line 202: Why not just insert all the rules in the OPENSHIFT chain here and skip the chain jump?

cgoncalves:
Does this tool also support IPv6?

types/types.go
yuvalk: because the input chain might already exist and have other stuff in it
or at least that was my initial thought when I saw this - that we have everything openshift need in it's own chain, and that can be used anywhere the user wants..
but then we might need a different way to modify the input chain?

sabinaaledort:
> Does this tool also support IPv6?

Yes but we actually wanted to split the communication matrix into two files, one for ipv4 ports and another for ipv6 ports. We found out there are no IPv6 endpointslices all are IPv4 due to the following bug https://issues.redhat.com/browse/OCPBUGS-35025

types/types.go
cgoncalves: I know. The thing is another entity manages the whole table so we should not touch it. There is even a warning comment in it. That is another advantage of nftables compared to iptables: you can have multiple tables. Also important is if two tables have chains on the same Netfilter hook, it is paramount to have them with different priorities or otherwise there is no guarantee of evaluation order:

> It's possible to give two base chains the same priority, but there is no guaranteed evaluation order of base chains with identical priority that are attached to the same hook location

https://wiki.nftables.org/wiki-nftables/index.php/Configuring_chains

In the latest PR change, Amal addressed my comment about the priority and table separation where the communication matrix rules have lower priority than other ones. If it had higher or same but nftables would evaluate those rules first, traffic could be dropped (there is a drop-all rule with log at the end of the proposed chain).

cgoncalves: Yuval, think of firewalld. It creates its own table so that it doesn't conflict with anything else on the system.

cgoncalves:
I had a chat with @aabughosh offline. She has now consolidated the nftables output in a single chain and in its own table. Additionally, she also added support for IPv6 by changing the table family to inet and adding icmpv6. I haven't tested with these latest changes but I like the way it is now.

cmd/main.go
```diff
@@ -79,136 +57,37 @@ func main() {
...
+		panic(fmt.Errorf("failed to create the communication matrix: %s", err))
 	}
-	defer udpFile.Close()
-
-	nodesList, err := cs.CoreV1Interface.Nodes().List(context.TODO(), metav1.ListOptions{})
+	// write the endpoint slice matrix to file
+	err = commatrix.WriteMatrixToFileByType(*mat, "communication-matrix", format, deployment, destDir)
 	if err != nil {
-		panic(err)
+		panic(fmt.Sprintf("Error while the endpoint slice matrix to file :%v", err))
```
sabinaaledort, line 68: Typo: missing `writing`

commatrix/generate.go
```diff
@@ -0,0 +1,207 @@
...
+}
+
+func WriteSSRawFiles(destDir string, ssOutTCP, ssOutUDP []byte) error {
+	tcpFile, udpFile, err := createSSOutputFiles(destDir)
+	if err != nil {
+		return err
+	}
+	defer tcpFile.Close()
+	defer udpFile.Close()
+	_, err = tcpFile.Write([]byte(string(ssOutTCP)))
```
sabinaaledort, line 122: nit: missing new line

commatrix/generate.go
```diff
@@ -0,0 +1,207 @@
...
+	if err != nil {
+		tcpFile.Close()
+		return nil, nil, err
+	}
+
+	return tcpFile, udpFile, nil
+}
+
+func WriteSSRawFiles(destDir string, ssOutTCP, ssOutUDP []byte) error {
+	tcpFile, udpFile, err := createSSOutputFiles(destDir)
```
sabinaaledort, line 116: I still think createSSOutputFiles is not needed and confusing, it's not really creating them, since the generate func is the one creating them, it's only opening the files

commatrix/generate.go
```diff
@@ -0,0 +1,207 @@
...
+	if err != nil {
+		return err
+	}
+	defer tcpFile.Close()
+	defer udpFile.Close()
+	_, err = tcpFile.Write([]byte(string(ssOutTCP)))
+	if err != nil {
+		return fmt.Errorf("failed writing to file: %s", err)
+	}
+	_, err = udpFile.Write([]byte(string(ssOutUDP)))
```
sabinaaledort, line 126: nit: missing new line

commatrix/generate.go
```diff
@@ -0,0 +1,207 @@
...
+
+	return nil
+}
+
+func WriteMatrixToFileByType(mat types.ComMatrix, fileNamePrefix, format string, deployment Deployment, destDir string) error {
+	printFn, err := getPrintFunction(format)
+	if err != nil {
+		return err
+	}
+	if format == types.FormatNFT {
```
sabinaaledort, line 139: nit: missing new line

commatrix/generate.go
```diff
@@ -0,0 +1,207 @@
...
+				return err
+			}
+		}
+	} else {
+		err := writeMatrixToFile(mat, fileNamePrefix, format, printFn, destDir)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
```
sabinaaledort, line 157: nit: missing new line

commatrix/commatrix_test.go
```diff
@@ -0,0 +1,52 @@
...
+			{NodeRole: "master", Service: "testService"},
+		},
+	}
+	printFn := types.ToJSON
+	fileName := "test-matrix"
+	format := "json"
+
+	err := writeMatrixToFile(matrix, fileName, format, printFn, destDir)
+	assert.NoError(t, err)
+	assert.FileExists(t, filepath.Join(destDir, "test-matrix.json"))
```
oribon, line 51: this doesn't do much, wdyt about expanding the test to a table (with all formats) and actually checking the content of the files? you can probably use what we did on metallb with the .golden files.

cmd/main.go
```diff
@@ -79,136 +57,37 @@ func main() {
...
-		})
+		panic(fmt.Sprintf("Error while generating the ss matrix and ss raws :%v", err))
 	}
-
-	err = g.Wait()
+	// write ss raw files
+	err = commatrix.WriteSSRawFiles(destDir, ssOutTCP, ssOutUDP)
 	if err != nil {
-		panic(err)
+		panic(fmt.Sprintf("Error while write the ss raw files :%v", err))
```
sabinaaledort, line 78: nit: write -> writing

PR #17, opened 2024-07-16
Added ComMatrix method; Changed private func to public

shirmoran:
This PR changed functions in order to use them in a new [openshift\origin](https://github.com/openshift/origin) test specified [here](https://issues.redhat.com/browse/CNF-12146) : 
Changed "addFromFile" function (in types.go) to a public function to allow using it in the mentioned above test.

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -160,36 +160,3 @@ func (m ComMatrix) Contains(cd ComDetails) bool {
 
 	return false
 }
-
-// Returns True if both ComMatrixes have exactly the same ComMatrix details.
-func (m ComMatrix) Equals(other ComMatrix) bool {
-	// Check if all of "m" commatrix details are in "other" commartix
-	for _, cd1 := range m.Matrix {
-		found := false
-		for _, cd2 := range other.Matrix {
-			if cd1.Equals(cd2) {
-				found = true
-				break
-			}
-		}
-		if !found {
-			return false
-		}
-	}
-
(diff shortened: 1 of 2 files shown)


types/types.go
```diff
@@ -160,3 +160,36 @@ func (m ComMatrix) Contains(cd ComDetails) bool {
 
 	return false
 }
+
+// Returns True if both ComMatrixes have exactly the same ComMatrix details.
+func (m ComMatrix) Equals(other ComMatrix) bool {
```
aabughosh, line 165: why not to use the Diff function?
u need from diff to be equal to nothing from 2 sides

aabughosh: @sabinaaledort also can use that on the test of the origin

shirmoran: Because we don't need the Diff matrices is this case, it occupies unnecessary space.

sabinaaledort:
I'm not sure the right thing to do is to keep the equals func in the commatrix repo instead of just use in the test.
It's an unused func in the library and makes an extra dependence between the test and the lib, for example if we want to ignore some ports in the test diff we will have to make a change in the matrix repo as well

aabughosh (approved):

sabinaaledort:
I think using addFromFile func for the test is an interesting idea but I think you should consider to compare the csv files directly, it's an an extra dependence that might cause bugs. For example if we have a bug in the addFromFile func and we miss a port the test will fail...

PR #20, opened 2024-07-28
creating a unit test using mocking for the generate SS function

aabughosh:
update the function: remove unused vars
and update the debug object to use the mock functions on test


debug/debug.go
```diff
@@ -17,20 +17,32 @@ import (
 	"github.com/openshift-kni/commatrix/client"
 )
 
+type DebugPodInterface interface {
```
sebrandon1, line 20: I would argue that more things in Go need to pass around interfaces instead of raw structs to allow for better testing in the future (if needed).  This seems fine to me.

sebrandon1 (approved):
Seems to be okay as far as unit tests go.  The only functional changes just move parameters from struct to interface which is fine.

commatrix/commatrix_test.go
```diff
@@ -50,3 +61,72 @@ func TestWriteMatrixToFile(t *testing.T) {
...
+
+	mockDebugPod.EXPECT().Clean().Return(nil).AnyTimes()
+
+	mockDebugPod.EXPECT().GetNodeName().Return("test-node").AnyTimes()
+
+	originalNew := debug.NewDebugPod
+	debug.NewDebugPod = func(cs *clientutil.ClientSet, node string, namespace string, image string) (debug.DebugPodInterface, error) {
+		return mockDebugPod, nil
+	}
+	defer func() { debug.NewDebugPod = originalNew }()
```
sebrandon1, line 108: Another way of doing it, is we could just create a func `func NewFakeDebugPod() DebugPod` within this file.

Since we passing around interfaces now, you shouldn't have to spoof `New()` at all.  Just setup your `DebugPod` struct with everything it needs before passing it along to the funcs that accept a `DebugPodInterface`.

sebrandon1: This would leave `New()` untouched and you wouldn't need `var NewDebugPod = New` in the `debug/debug.go` file.

PR #21, opened 2024-07-28
Allow generating commatrix from csv and fix samples

shirmoran:
The node role header of a commatrix in the csv format, defers from those of types json and yaml (`Node Role` instead of `nodeRole`). So, csv annotations in `ComDetails` struct were added.

The previous custom entries examples had some incorrect parameters which were fixed:

1. example-custom-entries.csv: `NodeRole` column was changed to `Node Rule` (as in the documented commatrix.
2. example-custom-entries.yaml and example-custom-entries.json: port was changed from a string to an int (9050 instead of "9050" for example).

a unit test applying the custom entries samples will be added in a separate PR.

### DIFF
--- a/commatrix/generate.go
+++ b/commatrix/generate.go
@@ -209,7 +209,7 @@ func getComMatrixHeadersByFormat(format string) (string, error) {
 		field := typ.Field(i)
 		tag := field.Tag.Get(format)
 		if tag == "" {
-			return "", fmt.Errorf("field %v has not tag of format %s", field, format)
+			return "", fmt.Errorf("field %v has no tag of format %s", field, format)
 		}
 		tagsList = append(tagsList, tag)
 	}
--- a/commatrix/generate.go
+++ b/commatrix/generate.go
@@ -164,15 +164,19 @@ func writeMatrixToFile(matrix types.ComMatrix, fileName, format string, printFn 
 	return os.WriteFile(comMatrixFileName, res, 0644)
 }
 
-func GenerateMatrixDiff(mat1 types.ComMatrix, mat2 types.ComMatrix) string {
-	diff := consts.CSVHeaders + "\n"
+// GenerateMatrixDiff generates a new matrix demonstrating the diff between mat2 to mat1
+func GenerateMatrixDiff(mat1 types.ComMatrix, mat2 types.ComMatrix) types.ComMatrix {
+	diffMatrix := []types.ComDetails{}
 	for _, cd := range mat1.Matrix {
 		if mat2.Contains(cd) {
-			diff += fmt.Sprintf("%s\n", cd)
+			diffMatrix = append(diffMatrix, cd)
 			continue
 		}
 
-		diff += fmt.Sprintf("+ %s\n", cd)
+		// add "+" before cd's mat1 contains but mat2 doesn't
+		newCd := cd
+		newCd.Direction = fmt.Sprintf("+ %s", newCd.Direction)
+		diffMatrix = append(diffMatrix, newCd)
 	}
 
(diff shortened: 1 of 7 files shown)


types/types.go
```diff
@@ -34,15 +34,15 @@ type ComMatrix struct {
...
+	Port      int    `json:"port" yaml:"port" csv:"Port"`
+	Namespace string `json:"namespace" yaml:"namespace" csv:"Namespace"`
+	Service   string `json:"service" yaml:"service" csv:"Service"`
+	Pod       string `json:"pod" yaml:"pod" csv:"Pod"`
+	Container string `json:"container" yaml:"container" csv:"Container"`
+	NodeRole  string `json:"nodeRole" yaml:"nodeRole" csv:"Node Role"`
+	Optional  bool   `json:"optional" yaml:"optional" csv:"Optional"`
 }
 
 func ToCSV(m ComMatrix) ([]byte, error) {
```
sabinaaledort, line 48: With this solution, should this func be updated and can we remove consts.CSVHeaders?

shirmoran: I'm not sure, since this func converts a matrix to CSV and the solution allows converting a CSV to a matrix.

commatrix/generate.go
```diff
@@ -164,14 +166,22 @@ func writeMatrixToFile(matrix types.ComMatrix, fileName, format string, printFn
 	return os.WriteFile(comMatrixFileName, res, 0644)
 }
 
-func GenerateMatrixDiff(mat1 types.ComMatrix, mat2 types.ComMatrix) string {
-	diff := consts.CSVHeaders + "\n"
+// GenerateMatrixDiff generates a new matrix demonstrating the diff between mat1 to mat2.
```
sabinaaledort, line 169: I would avoid saying it's generates a matrix, it's generates a diff file

commatrix/generate.go
```diff
@@ -183,11 +193,27 @@ func GenerateMatrixDiff(mat1 types.ComMatrix, mat2 types.ComMatrix) string {
...
+			// add "-" before cd's mat1 doesn't contain but mat2 does
 			diff += fmt.Sprintf("- %s\n", cd)
 		}
 	}
 
-	return diff
+	return diff, nil
+}
+
+func getComMatrixColTagsByFormat(format string) string {
```
sabinaaledort, line 204: I'm not sure about the name of the func, it says "get cols" while cols are specific to csv and it can generate the headers of a json/yaml formats as well where these are not columns

shirmoran: I agree, maybe `getComMatrixHeaderTagsByFormat` is a better name. what do you think?

sabinaaledort: getComMatrixHeadersByFormat makes sense to me

commatrix/generate.go
```diff
@@ -164,14 +166,22 @@ func writeMatrixToFile(matrix types.ComMatrix, fileName, format string, printFn
 	return os.WriteFile(comMatrixFileName, res, 0644)
 }
 
-func GenerateMatrixDiff(mat1 types.ComMatrix, mat2 types.ComMatrix) string {
-	diff := consts.CSVHeaders + "\n"
+// GenerateMatrixDiff generates the diff between mat1 to mat2.
+func GenerateMatrixDiff(mat1 types.ComMatrix, mat2 types.ComMatrix) (string, error) {
```
sebrandon1, line 170:
```suggestion
func GenerateMatrixDiff(mat1, mat2 types.ComMatrix) (string, error) {
```

sebrandon1 (commented):
Minor comment about params

aabughosh (approved):

PR #22, opened 2024-08-04
Refactor project

SchSeba:
TODOS:

- [ ] Add debug logs when running the code with the --debug flag
- [ ] Add unit tests


SchSeba:
Hi @sabinaaledort @yuvalk @shirmoran @aabughosh please take a look on this refactor.

I will for the conflicts tomorrow morning

pkg/types/types.go
```diff
@@ -0,0 +1,377 @@
...
+		if tag == "" {
+			return "", fmt.Errorf("field %v has no tag of format %s", field, format)
+		}
+		tagsList = append(tagsList, tag)
+	}
+
+	return strings.Join(tagsList, ","), nil
+}
+
+func CleanComDetails(outPuts []ComDetails) []ComDetails {
```
shirmoran, line 338: Suggestion: Changing this func to a method of  `ComMatrix`, that applies the `ComDetails` list changes on the matrix instead of returning it. 
What do you think?

aabughosh: yah i agree with shir since the [] ComDetails is a ComMatrix

pkg/connections/connections.go
```diff
@@ -0,0 +1,331 @@
...
+		format,
+		destDir,
+		env,
+		deployment,
+		nodesRolesMap,
+	}, nil
+}
+
+func (cc *ConnectionCheck) GenerateSS() (*types.ComMatrix, []byte, []byte, error) {
+	nodesList := &corev1.NodeList{}
```
aabughosh, line 69: why not to save the node names on the ConnectionCheck? 
as the node role list u can also save the node name and we will not use the corv1 to list

pkg/types/types.go
```diff
@@ -0,0 +1,377 @@
...
+)
+
+type Deployment int
+
+const (
+	SNO Deployment = iota
+	MNO
+)
+
+type Format int
```
shirmoran, line 37: Is this type and following consts necessary? Seems like we only use the string format

cmd/main.go
```diff
@@ -25,71 +32,83 @@ func init() {
 	flag.StringVar(&deploymentStr, "deployment", "mno", "Deployment type (mno/sno)")
 	flag.StringVar(&customEntriesPath, "customEntriesPath", "", "Add custom entries from a file to the matrix")
 	flag.StringVar(&customEntriesFormat, "customEntriesFormat", "", "Set the format of the custom entries file (json,yaml,csv)")
+	flag.BoolVar(&debug, "debug", false, "Debug logs")
```
sabinaaledort, line 35: Should this flag be configurable in the Makefile?

pkg/types/types.go
aabughosh: we are using the strings and this we are not using i think this we can remove

pkg/connections/connections.go
```diff
@@ -0,0 +1,331 @@
...
+
+	for _, ssEntry := range ssOutput {
+		cd := parseComDetail(ssEntry)
+
+		name, err := cc.getContainerName(debugPod, ssEntry)
+		if err != nil {
+			log.Debugf("failed to identify container for ss entry: %serr: %s", ssEntry, err)
+		}
+
+		for _, role := range nodeRole {
```
sabinaaledort, line 196: I think in SNO we should have one role only - master, to avoid duplicating the matrix

cmd/main.go
```diff
@@ -25,71 +32,83 @@ func init() {
...
+
+	commMatrix, err := commatrixcreator.New(epExporter, customEntriesPath, customEntriesFormat, env, deployment)
 	if err != nil {
-		panic(fmt.Errorf("failed to create the communication matrix: %s", err))
+		panic(fmt.Errorf("failed creating comm matrix: %w", err))
 	}
-	// write the endpoint slice matrix to file
-	err = commatrix.WriteMatrixToFileByType(*mat, "communication-matrix", format, deployment, destDir)
+
+	matrix, err := commMatrix.CreateConnectionMatrix()
```
sabinaaledort, line 75: How about naming the connections `listening sockets`?

pkg/connections/connections.go
```diff
@@ -0,0 +1,331 @@
...
+
+	_, err = udpFile.Write(ssOutUDP)
+	if err != nil {
+		return fmt.Errorf("failed writing to file: %s", err)
+	}
+
+	return nil
+}
+
+func (cc *ConnectionCheck) createSSOutputFromNode(debugPod *corev1.Pod, node *corev1.Node) ([]types.ComDetails, []byte, []byte, error) {
```
aabughosh, line 157: u can send node name and not the  node *corev1.Node
since we are just using name

pkg/connections/connections.go
aabughosh: i think we are are calling CleanComDetails after the toComDetails and with this we are avoiding the duplicates

pkg/connections/connections.go
SchSeba: I like the idea nice!

pkg/connections/connections.go
SchSeba: lets use the first role for now and we can change that in a following PR

pkg/types/types.go
SchSeba: yep

pkg/types/types.go
SchSeba: sounds good :)

pkg/listening-sockets/listening-sockets.go
```diff
@@ -0,0 +1,309 @@
...
+		err := cc.podUtils.DeleteNamespace(consts.DefaultDebugNamespace)
+		if err != nil {
+			fmt.Printf("failed to delete debug namespace: %v", err)
+		}
+	}()
+	var ssOutTCP, ssOutUDP []byte
+	nLock := &sync.Mutex{}
+	g := new(errgroup.Group)
+	for nodeName := range cc.nodesRolesMap {
+		name := nodeName
```
aabughosh, line 84: why u are doing that and not stay with the nodeName?

pkg/types/types.go
```diff
@@ -0,0 +1,368 @@
...
+			diff += fmt.Sprintf("- %s\n", cd)
+		}
+	}
+
+	return diff, nil
+}
+
+func (m *ComMatrix) print(format string) ([]byte, error) {
+	switch format {
+	case "json":
```
aabughosh, line 194: u can here use the 

	FormatJSON = "json"
	FormatYAML = "yaml"
	FormatCSV  = "csv"
	FormatNFT  = "nft"

pkg/listening-sockets/listening-sockets.go
SchSeba: you can't because you use that in a go routine

pkg/types/types.go
SchSeba: right!

OWNERS_ALIASES
```diff
@@ -0,0 +1,13 @@
+aliases:
+  cnf-features-reviewers:
```
sabinaaledort, line 2: cnf-features-reviewers/approvers -> commatrix-reviewers/approvers

SchSeba: done

PR #25, opened 2024-08-06
Fix GetNodeRole func to return a single node role per node

sabinaaledort:
We should return a single node role per node to avoid duplicating the matrix with the same ports on different node roles while listening on the same node (for example, in SNO, which has multiple roles on a single node).

### DIFF
--- a/pkg/endpointslices/endpointslices.go
+++ b/pkg/endpointslices/endpointslices.go
@@ -25,8 +25,8 @@ type EndpointSlicesInfo struct {
 
 type EndpointSlicesExporter struct {
 	*client.ClientSet
-	nodesRolesMap map[string][]string
-	sliceInfo     []EndpointSlicesInfo
+	nodeToRole map[string]string
+	sliceInfo  []EndpointSlicesInfo
 }
 
 func New(cs *client.ClientSet) (*EndpointSlicesExporter, error) {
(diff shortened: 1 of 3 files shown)


pkg/endpointslices/endpointslices.go
```diff
@@ -138,7 +136,7 @@ func (ei *EndpointSlicesInfo) getEndpointSliceNodeRoles(nodeRoles map[string][]s
 	return roles
 }
 
-func (ei *EndpointSlicesInfo) toComDetails(nodeRoles map[string][]string) ([]types.ComDetails, error) {
+func (ei *EndpointSlicesInfo) toComDetails(nodesRoles map[string]string) ([]types.ComDetails, error) {
```
SchSeba, line 139: nodesRoles -> nodesRole

pkg/listening-sockets/listening-sockets.go
```diff
@@ -36,7 +36,7 @@ type ConnectionCheck struct {
 	destDir             string
 	env                 types.Env
 	deployment          types.Deployment
-	nodesRolesMap       map[string][]string
+	nodesRolesMap       map[string]string
```
SchSeba, line 39: nodesRoleMap

SchSeba (commented):
just small nits and we can merge

pkg/endpointslices/endpointslices.go
```diff
@@ -35,9 +35,9 @@ func New(cs *client.ClientSet) (*EndpointSlicesExporter, error) {
 	if err != nil {
 		return nil, err
 	}
-	nodesRolesMap := map[string][]string{}
+	nodesRolesMap := map[string]string{}
 	for _, node := range nodeList.Items {
-		nodesRolesMap[node.Name] = types.GetNodeRoles(&node)
+		nodesRolesMap[node.Name] = types.GetNodeRole(&node)
```
aabughosh, line 40: we dont need to check the output if it not ""?

shirmoran: I also think we should check the output is not "".

aabughosh (approved):

oribon (commented):
lgtm

PR #31, opened 2024-08-08
add log implementation and add debug logs on debug mod

aabughosh:


SchSeba:
lgtm

please wait for https://github.com/openshift-kni/commatrix/pull/34 and rebase so we can have a CI run

SchSeba (approved):
/lgtm

SchSeba:
/hold

cc @sabinaaledort

pkg/commatrix-creator/commatrix.go
```diff
@@ -39,97 +40,125 @@ func New(exporter *endpointslices.EndpointSlicesExporter, customEntriesPath stri
...
+	log := utils.Logger() // Get the logger instance from utils
+
+	log.Debug("Determining static entries based on environment and deployment")
 	comDetails := []types.ComDetails{}
 
 	switch cm.e {
 	case types.Baremetal:
+		log.Debug("Adding Baremetal static entries")
 		comDetails = append(comDetails, types.BaremetalStaticEntriesMaster...)
-		if cm.d == types.SNO {
```
sabinaaledort, line 112: Why is this change?

aabughosh: mm because of the break i wrote on other way but its the same

aabughosh: comDetails = append(comDetails, types.CloudStaticEntriesMaster...)
		if cm.d == types.SNO {
			break
			}
			comDetails = append(comDetails, types.CloudStaticEntriesWorker...)
	before we had that so i put itnsideon the if

sabinaaledort: The idea of checking with equals is to make it clear in the code that this is what has to be done for a specific case. If we add more cases in the future we will have to rollback to equals.

PR #32, opened 2024-08-11
Add new error to skip endpointslices with empty OwnerReferences

sabinaaledort:
Adding a new error to skip endpointslices with empty OwnerReferences instead of causing the entire matrix build to fail.

### DIFF
--- a/pkg/endpointslices/endpointslices.go
+++ b/pkg/endpointslices/endpointslices.go
@@ -29,6 +29,15 @@ type EndpointSlicesExporter struct {
 	sliceInfo  []EndpointSlicesInfo
 }
 
+type NoOwnerRefErr struct {
+	name      string
+	namespace string
+}
+
+func (e *NoOwnerRefErr) Error() string {
+	return fmt.Sprintf("empty OwnerReferences in EndpointSlice %s/%s. skipping", e.namespace, e.name)
+}
+
 func New(cs *client.ClientSet) (*EndpointSlicesExporter, error) {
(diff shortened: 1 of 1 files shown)


shirmoran (approved):

pkg/endpointslices/endpointslices.go
```diff
@@ -105,9 +115,13 @@ func (ep *EndpointSlicesExporter) ToComDetails() ([]types.ComDetails, error) {
 	for _, epSliceInfo := range ep.sliceInfo {
 		cds, err := epSliceInfo.toComDetails(ep.nodeToRole)
 		if err != nil {
+			var noOwnerRefErr *NoOwnerRefErr
```
SchSeba, line 118: just a nit move this one outside the for loop

pkg/endpointslices/endpointslices.go
```diff
@@ -29,6 +29,15 @@ type EndpointSlicesExporter struct {
 	sliceInfo  []EndpointSlicesInfo
 }
 
+type NoOwnerRefErr struct {
```
aabughosh, line 32: what do u think to add this to utils?
since its not EndpointSlice

SchSeba: I think we can leave it here as we don't use it in any other place and I don't want to make utils a kitchen sink

SchSeba (approved):
/lgtm
/approve

aabughosh (approved):

PR #35, opened 2024-08-12
Sort matrix diff entries order

shirmoran:
This PR modifies ComMatrix method `GenerateMatrixDiff` to generate an organized diff using the `CleanComDetails` method sorting logic.

The updated `GenerateMatrixDiff` method works as follows:
1. Calling `generateUnitedMatrix` method to unite two matrices (without duplications).
2. Sorting the United matrix using `CleanComDetails` method.
3. Calling `mapDiffBetweenMatrices` method mark the diff between the matrices using a map.
4. Iterating over the **sorted** United matrix and generate the diff:  a `+` or  `-` sign (or a nothing) will be added to the each cd according to the marking done in previous (3) step.

### DIFF
--- a/pkg/commatrix-creator/commatrix.go
+++ b/pkg/commatrix-creator/commatrix.go
@@ -73,7 +73,7 @@ func (cm *CommunicationMatrixCreator) CreateEndpointMatrix() (*types.ComMatrix, 
 	}
 
 	commMatrix := &types.ComMatrix{Matrix: epSliceComDetails}
-	log.Debug("Sorting ComMatrix and remove duplicates")
+	log.Debug("Sorting ComMatrix and removing duplicates")
 	commMatrix.SortAndRemoveDuplicates()
 	return commMatrix, nil
 }
--- a/pkg/commatrix-creator/commatrix.go
+++ b/pkg/commatrix-creator/commatrix.go
@@ -64,7 +64,7 @@ func (cm *CommunicationMatrixCreator) CreateEndpointMatrix() (*types.ComMatrix, 
 	}
 
 	commMatrix := &types.ComMatrix{Matrix: epSliceComDetails}
-	commMatrix.CleanComDetails()
+	commMatrix.SortAndRemoveDuplicates()
 	return commMatrix, nil
 }
 
--- a/pkg/types/types.go
+++ b/pkg/types/types.go
@@ -172,7 +172,7 @@ func (m *ComMatrix) generateUnitedMatrix(other *ComMatrix) *ComMatrix {
 // cd which m contains but other doesn't --> 1
 // cd both m and other contains --> 0
 // cd which other doesn't contain but m does --> -1.
-func (m *ComMatrix) mapDiffBetweenMatrices(other *ComMatrix) map[string]int {
+func (m *ComMatrix) markDiffBetweenMatrices(other *ComMatrix) map[string]int {
 	mapComDetailToSign := make(map[string]int)
 
 	for _, cd := range m.Matrix {
@@ -204,7 +204,7 @@ func (m *ComMatrix) mapDiffBetweenMatrices(other *ComMatrix) map[string]int {
 func (m *ComMatrix) GenerateMatrixDiff(other *ComMatrix) (string, error) {
 	unitedComMatrix := m.generateUnitedMatrix(other)
 	unitedComMatrix.CleanComDetails()
-	mapComDetailToSign := m.mapDiffBetweenMatrices(other)
+	mapComDetailToSign := m.markDiffBetweenMatrices(other)
 
 	colNames, err := getComMatrixHeadersByFormat(FormatCSV)
 	if err != nil {
--- a/pkg/types/types.go
+++ b/pkg/types/types.go
@@ -154,21 +154,25 @@ func (m *ComMatrix) WriteMatrixToFileByType(utilsHelpers utils.UtilsInterface, f
 	return nil
 }
 
-// GenerateUnitedMatrix generates a new matrix from m and other with no duplications.
-func (m *ComMatrix) generateUnitedMatrix(other *ComMatrix) *ComMatrix {
-	// initial united matrix with m.Matrix
-	unitedComDetails := m.Matrix
+// Combine generates a new sorted matrix from m and other with no duplications.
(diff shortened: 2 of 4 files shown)


pkg/types/types.go
```diff
@@ -154,35 +154,75 @@ func (m *ComMatrix) WriteMatrixToFileByType(utilsHelpers utils.UtilsInterface, f
...
+			// m doesn't contain cd but other does
+			mapComDetailToSign[cd.String()] = -1
+		}
+	}
+
+	return mapComDetailToSign
+}
+
+// GenerateMatrixDiff generates the diff between mat1 to mat2.
+func (m *ComMatrix) GenerateMatrixDiff(other *ComMatrix) (string, error) {
```
aabughosh, line 204: maybe need to change the name?
its generating a string

shirmoran: `GenerateDiff` works in my opinion. @sabinaaledort @aabughosh what do you think?

pkg/types/types.go
```diff
@@ -154,35 +154,78 @@ func (m *ComMatrix) WriteMatrixToFileByType(utilsHelpers utils.UtilsInterface, f
...
-
-		// add "+" before cd's mat1 contains but mat2 doesn't
-		diff += fmt.Sprintf("+ %s\n", cd)
 	}
 
-	for _, cd := range mat2.Matrix {
+	for _, cd := range other.Matrix {
 		// Skip "rpc.statd" ports, these are randomly open ports on the node,
 		// no need to mention them in the matrix diff
 		if cd.Service == "rpc.statd" {
```
sabinaaledort, line 194: I'm not sure if this is the right place to make this check

shirmoran: As we discussed, the ss filter should be only in the origin, or am I wrong? 
Maybe we can address this subject in a separate PR?

pkg/types/types.go
```diff
@@ -154,35 +154,78 @@ func (m *ComMatrix) WriteMatrixToFileByType(utilsHelpers utils.UtilsInterface, f
 	return nil
 }
 
-// GenerateMatrixDiff generates the diff between mat1 to mat2.
-func (m *ComMatrix) GenerateMatrixDiff(mat2 *ComMatrix) (string, error) {
-	colNames, err := getComMatrixHeadersByFormat(FormatCSV)
-	if err != nil {
-		return "", fmt.Errorf("error getting commatrix CSV tags: %v", err)
+// Combine generates a new sorted matrix from m and other with no duplications.
```
sabinaaledort, line 157: I think if call it add, add the two together and do the clean (remove dups and sort) is will do the same with less changes?

pkg/types/types.go
```diff
@@ -154,35 +154,78 @@ func (m *ComMatrix) WriteMatrixToFileByType(utilsHelpers utils.UtilsInterface, f
...
+			mapComDetailToSign[cd.String()] = -1
+		}
+	}
+
+	return mapComDetailToSign
+}
+
+// GenerateMatrixDiff generates the diff between mat1 to mat2.
+func (m *ComMatrix) GenerateMatrixDiff(other *ComMatrix) (string, error) {
+	combinedComMatrix := m.combine(other)
```
sabinaaledort, line 209: Maybe call it diff matrix?

pkg/types/types.go
shirmoran: you mean adding the matrix at the end without removing the duplicates, and than call the clean function?

sabinaaledort: As part of the generate you can add the two mats with a single append and call clean.

shirmoran: okay yeah. Though maybe we can change the name of the method `CleanComDetails` to `SortAndRemoveDuplicates`, I think it is more clear that way. what do you think?

sabinaaledort: Totally agree

pkg/commatrix-creator/commatrix.go
```diff
@@ -73,8 +73,8 @@ func (cm *CommunicationMatrixCreator) CreateEndpointMatrix() (*types.ComMatrix,
 	}
 
 	commMatrix := &types.ComMatrix{Matrix: epSliceComDetails}
-	log.Debug("Cleaning ComDetails")
-	commMatrix.CleanComDetails()
+	log.Debug("Sorting ComMatrix and remove duplicates")
```
sabinaaledort, line 76: nit: sorting and removing / sort and remove

pkg/types/types.go
```diff
@@ -154,35 +154,65 @@ func (m *ComMatrix) WriteMatrixToFileByType(utilsHelpers utils.UtilsInterface, f
...
+			mapComDetailToSign[cd.String()] = -1
+		}
+	}
+
+	return mapComDetailToSign
+}
+
+// GenerateMatrixDiff generates the diff between mat1 to mat2.
+func (m *ComMatrix) GenerateDiff(other *ComMatrix) (string, error) {
+	combinedComDetails := append([]ComDetails{}, m.Matrix...)
```
sabinaaledort, line 191: combinedComDetails := append(m.Matrix, other.Matrix...)

pkg/types/types.go
```diff
@@ -299,7 +329,8 @@ func (m *ComMatrix) ToNFTables() ([]byte, error) {
 	return []byte(result), nil
 }
 
-func (m *ComMatrix) CleanComDetails() {
+// CleanComDetails deletes duplicates in matrix and sort it.
```
sabinaaledort, line 332: Fix comment

pkg/types/types.go
shirmoran: I'm getting a linting error when doing so.

shirmoran: `appendAssign: append result not assigned to the same slice (gocritic)`

sabinaaledort: ok so you can do
```
combindComDetails := m.Matrix
combinedComDetails = append(combinedComDetails, other.Matrix...)
```

shirmoran: That way it shallow copies the matrix since it is a slice, I think it might be better to deep copy as I did so the original won't change as we iterate through it for mapping the sign. what do you think?

sabinaaledort: I think for our goal it's sufficient but let's keep it as it is

PR #36, opened 2024-08-13
add the firewall apply test to e2e tests and add it to the make file + add reboot test

aabughosh:
ginkgo test that apply firewall by blocking all ports except the ones OCP is listening on 
and reboot the node and verify that the firewall rules persist after the reboot.


test/e2e/e2e_tests_suite_test.go
```diff
@@ -0,0 +1,221 @@
...
+
+		command = []string{"bash", "-c",
+			"systemctl enable nftables",
+		}
+		_, err = utilsHelpers.RunCommandOnPod(debugPod, command)
+		if err != nil {
+			return fmt.Errorf("failed to enable nftables %q on node %s: %w", command, node.Name, err)
+		}
+
+		artifactsDir, _ := os.LookupEnv("ARTIFACT_DIR")
```
sabinaaledort, line 178: Should check the returned bool

test/e2e/e2e_tests_suite_test.go
```diff
@@ -0,0 +1,221 @@
...
+	return nil
+}
+
+// isSNOCluster will check if OCP is a single node cluster.
+func isSNOCluster(cs *client.ClientSet) (bool, error) {
+	nodes, err := cs.CoreV1Interface.Nodes().List(context.Background(), metav1.ListOptions{})
+	if err != nil {
+		return false, err
+	}
+	if len(nodes.Items) > 1 {
```
sabinaaledort, line 200: Finding one node can also mean the cluster is unhealthy (MNO with broken node), better check the cluster infrastructure status

test/e2e/e2e_tests_suite_test.go
```diff
@@ -0,0 +1,221 @@
...
+		return false, nil
+	}
+	return true, nil
+}
+
+func infrtType() (bool, error) {
```
sabinaaledort, line 206: Can you add some spaces to make it more readable?

test/e2e/e2e_tests_suite_test.go
```diff
@@ -0,0 +1,221 @@
...
+		}
+		var cleanupErr error
+		defer func() {
+			if err := utilsHelpers.DeleteNamespace(ns); err != nil {
+				cleanupErr = err
+			}
+		}()
+
+		if cleanupErr != nil {
+			// Handle the error (log it, fail the test, etc.)
```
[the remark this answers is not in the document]
sabinaaledort: If cleanupErr is only set in the defer that won't do anything, defer is happening in the end of the run

test/e2e/e2e_tests_suite_test.go
```diff
@@ -0,0 +1,221 @@
...
+			return fmt.Errorf("failed to apply rule set %q on node %s: %w", command, node.Name, err)
+		}
+
+		// Check the output of nft list ruleset
+		command = []string{"bash", "-c", "/usr/sbin/nft list ruleset"}
+		output, err := utilsHelpers.RunCommandOnPod(debugPod, command)
+		if err != nil {
+			return fmt.Errorf("failed to list NFT ruleset on node %s: %w", node.Name, err)
+		}
+		fmt.Printf("Command output: %s\n", string(output))
```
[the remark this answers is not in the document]
aabughosh: no i will delete

test/e2e/e2e_tests_suite_test.go
```diff
@@ -0,0 +1,221 @@
...
+		command := []string{
+			"bash", "-c",
+			fmt.Sprintf("mkdir -p nft && echo '%s' > nft/firewall.nft", string(NFTtable)),
+		}
+		_, err = utilsHelpers.RunCommandOnPod(debugPod, command)
+		if err != nil {
+			return fmt.Errorf("failed to save rule set to file on node %s: %w", node.Name, err)
+		}
+
+		installCmd := []string{"bash", "-c", "yum update -y && yum install -y nftables"}
```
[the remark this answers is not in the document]
aabughosh: i asked carlos he told me this is fine

test/e2e/e2e_tests_suite_test.go
aabughosh: i did that like that becouse of local test i will add a print if not set but on the artifacts that should set

test/e2e/e2e_tests_suite_test.go
```diff
@@ -0,0 +1,222 @@
+package e2e
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	configv1 "github.com/openshift/api/config/v1"
```
SchSeba, line 9: please order the imports like this

```
	"context"
	"fmt"
	"os"
	"path/filepath"
        
	. "github.com/onsi/ginkgo/v2"
	. "github.com/onsi/gomega"

	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
	"k8s.io/client-go/tools/clientcmd"

	configv1 "github.com/openshift/api/config/v1"
	configv1client "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"

	"github.com/openshift-kni/commatrix/pkg/consts"
	"github.com/openshift-kni/commatrix/pkg/types"
	"github.com/openshift-kni/commatrix/pkg/utils"
        "github.com/openshift-kni/commatrix/pkg/client"
	commatrixcreator "github.com/openshift-kni/commatrix/pkg/commatrix-creator"
	"github.com/openshift-kni/commatrix/pkg/endpointslices"


```

test/e2e/e2e_tests_suite_test.go
```diff
@@ -0,0 +1,222 @@
...
+var _ = Describe("commatrix", func() {
+	It("should apply firewall by blocking all ports except the ones OCP is listening on", func() {
+		By("generating the commatrix")
+
+		kubeconfig, ok := os.LookupEnv("KUBECONFIG")
+		if !ok {
+			Fail("must set the KUBECONFIG environment variable")
+		}
+
+		config, err := clientcmd.BuildConfigFromFlags("", kubeconfig)
```
SchSeba, line 36: please move all of this to BeforeSuite take look on cnf-feature-deploy repo for examples

test/e2e/e2e_tests_suite_test.go
```diff
@@ -0,0 +1,222 @@
...
+		Expect(err).NotTo(HaveOccurred())
+
+		oc := configv1client.NewForConfigOrDie(config)
```
SchSeba, line 39: don't create a specific client just add what you need to the general client in the project

test/e2e/e2e_tests_suite_test.go
```diff
@@ -0,0 +1,222 @@
...
+			deployment = types.SNO
+		}
+
+		infra := types.Cloud
+		isBM, err := infrType(oc)
+		Expect(err).NotTo(HaveOccurred())
+
+		if isBM {
+			infra = types.Baremetal
+		}
```
SchSeba, line 58: This needs to be in the initial state not part of the test

test/e2e/e2e_tests_suite_test.go
```diff
@@ -0,0 +1,222 @@
...
+
+		if !isSNO {
+			By("apply firewall on worker nodes")
+
+			workerNFT, err := workerMat.ToNFTables()
+			Expect(err).NotTo(HaveOccurred())
+
+			err = applyRulesToNode(cs, "worker", workerNFT)
+			Expect(err).ToNot(HaveOccurred())
+		}
```
SchSeba, line 87: all of this needs to be in a before all or before each and you should put in the It a test for example to check the rules or something like that

test/e2e/e2e_tests_suite_test.go
```diff
@@ -0,0 +1,222 @@
...
+	})
+})
+
+func applyRulesToNode(cs *client.ClientSet, role string, NFTtable []byte) error {
+	utilsHelpers := utils.New(cs)
+
+	nodesList, err := cs.CoreV1Interface.Nodes().List(context.TODO(), metav1.ListOptions{})
```
SchSeba, line 94: please use the generic client like we have in the code itself

test/e2e/e2e_tests_suite_test.go
```diff
@@ -0,0 +1,222 @@
...
+	if err != nil {
```
SchSeba, line 95: this function is in the tests folder you can just do Expect(err).ToNot(HaveOccurred())

test/e2e/e2e_tests_suite_test.go
```diff
@@ -0,0 +1,222 @@
...
+			"bash", "-c",
+			fmt.Sprintf("mkdir -p nft && echo '%s' > nft/firewall.nft", string(NFTtable)),
+		}
+
+		_, err = utilsHelpers.RunCommandOnPod(debugPod, command)
+		if err != nil {
+			return fmt.Errorf("failed to save rule set to file on node %s: %w", node.Name, err)
+		}
+
+		installCmd := []string{"bash", "-c", "yum update -y && yum install -y nftables"}
```
SchSeba, line 143: this will not work on disconnected env and also it's a waste of time for the test multiple by the number of nodes (as we don't do this is parallel here)

test/e2e/e2e_tests_suite_test.go
```diff
@@ -0,0 +1,222 @@
...
+		_, err = utilsHelpers.RunCommandOnPod(debugPod, installCmd)
+		if err != nil {
+			return fmt.Errorf("failed to install nftables on  debugpod on node%s: %w", node.Name, err)
+		}
+
+		command = []string{
```
SchSeba, line 149: please create and helper function here for running a command we do the same multiple times.

you can also run everything in one command no need to exec the pod multiple times

test/e2e/e2e_tests_suite_test.go
```diff
@@ -0,0 +1,222 @@
...
+			"bash", "-c",
+			"/usr/sbin/nft list ruleset > /etc/nftables.conf",
+		}
+
+		_, err = utilsHelpers.RunCommandOnPod(debugPod, command)
+		if err != nil {
+			return fmt.Errorf("failed to save NFT ruleset to file on node %s: %w", node.Name, err)
+		}
+
+		command = []string{"bash", "-c",
```
SchSeba, line 178: do we need to enable the nftables service on the node?

test/e2e/e2e_tests_suite_test.go
```diff
@@ -0,0 +1,222 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to create artifactsDir %w", err)
+		}
+
+		// save the nft ruleset on ArtifactDirPath.
+		err = os.WriteFile(filepath.Join(comArtifactsDir, fmt.Sprintf("nftables-%s.nft", node.Name)), output, 0644)
+		if err != nil {
+			return fmt.Errorf("failed to save nft file on for node artifacts %s %w", node.Name, err)
+		}
+	}
```
SchSeba, line 201: this may be in a different function + as this is a test please add some test like checking rules are applied or something like that

test/e2e/e2e_tests_suite_test.go
```diff
@@ -0,0 +1,222 @@
...
+	infrastructureType, err := oc.Infrastructures().Get(context.Background(), "cluster", metav1.GetOptions{})
+	if err != nil {
+		return false, err
+	}
+
+	return infrastructureType.Status.ControlPlaneTopology == configv1.SingleReplicaTopologyMode, nil
+}
+
+func infrType(oc *configv1client.ConfigV1Client) (bool, error) {
+	infra, err := oc.Infrastructures().Get(context.Background(), "cluster", metav1.GetOptions{})
```
SchSeba, line 216: use generic client please

test/e2e/e2e_tests_suite_test.go
```diff
@@ -0,0 +1,222 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to save nft file on for node artifacts %s %w", node.Name, err)
+		}
+	}
+	return nil
+}
+
+// isSNOCluster will check if OCP is a single node cluster.
+func isSNOCluster(oc *configv1client.ConfigV1Client) (bool, error) {
+	infrastructureType, err := oc.Infrastructures().Get(context.Background(), "cluster", metav1.GetOptions{})
```
SchSeba, line 207: use generic client pleaseuse generic client please

Makefile
```diff
@@ -52,6 +52,10 @@ ginkgo:
 deps-update:
 	go mod tidy
 
+e2e-test:
+	go build -o apply_rules test/firewall_run.go 
+	./apply_rules
```
sabinaaledort, line 57: If we add more tests to the e2e suite and would like to focus on one the rules will always be applied and might break it

test/firewall/firewall.go
```diff
@@ -0,0 +1,233 @@
...
+func isSNOCluster(oc *configv1client.ConfigV1Client) (bool, error) {
+	infrastructureType, err := oc.Infrastructures().Get(context.Background(), "cluster", metav1.GetOptions{})
+	if err != nil {
+		return false, err
+	}
+
+	return infrastructureType.Status.ControlPlaneTopology == configv1.SingleReplicaTopologyMode, nil
+}
+
+func infraType(oc *configv1client.ConfigV1Client) (bool, error) {
```
sabinaaledort, line 226: The name of the func is `infraType` while it doesn't return the infra type, it's checking if it's BM and return bool

test/firewall/firewall.go
```diff
@@ -0,0 +1,233 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func Execute() error {
+	kubeconfig, ok := os.LookupEnv("KUBECONFIG")
+	if !ok {
+		return fmt.Errorf("must set the KUBECONFIG environment variable")
+	}
+
+	config, err := clientcmd.BuildConfigFromFlags("", kubeconfig)
```
sabinaaledort, line 31: You can create the client first and avoid `BuildConfigFromFlags`
```
cs, err := client.New()
configClient := configv1client.NewForConfigOrDie(cs.Config)
```

test/firewall_run.go
```diff
@@ -0,0 +1,18 @@
...
+package main
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/openshift-kni/commatrix/test/firewall"
+)
+
+func main() {
```
sabinaaledort, line 12: Why it that package main? Please use ginkgo standards

test/firewall/firewall.go
```diff
@@ -0,0 +1,233 @@
...
+	if err != nil {
+		return fmt.Errorf("error building kubeconfig: %w", err)
```
sabinaaledort, line 33: If that's a ginkgo test all the errors should be checked

test/firewall/firewall.go
```diff
@@ -0,0 +1,233 @@
...
+	masterNFT, err := masterMat.ToNFTables()
+	if err != nil {
+		return fmt.Errorf("error converting master matrix to NFTables: %w", err)
+	}
+
+	err = applyRulesToNode(cs, "master", masterNFT)
+	if err != nil {
+		return fmt.Errorf("error apply rules to master nodes: %w", err)
+	}
+	if !isSNO {
```
sabinaaledort, line 88: nit: missing new line

test/e2e/e2e_tests_suite_test.go
```diff
@@ -0,0 +1,142 @@
...
+		}
+
+	})
+
+	It("should apply firewall by blocking all ports except the ones OCP is listening on", func() {
+		By("apply firewall on nodes")
+		var NFTtable []byte
+		nodesList, err := cs.CoreV1Interface.Nodes().List(context.TODO(), metav1.ListOptions{})
+		Expect(err).ToNot(HaveOccurred())
+		By("nodesList nodesList")
```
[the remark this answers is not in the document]
aabughosh: yes

test/e2e/e2e_tests_suite_test.go
SchSeba: please put the imports in the right order

test/e2e/e2e_tests_suite_test.go
```diff
@@ -0,0 +1,161 @@
...
+
+var (
+	cs           *client.ClientSet
+	matrix       *types.ComMatrix
+	isSNO        bool
+	utilsHelpers utils.UtilsInterface
+)
+
+var _ = Describe("commatrix", func() {
+	BeforeEach(func() {
```
SchSeba, line 36: do this in the beforeSuite

test/e2e/e2e_tests_suite_test.go
```diff
@@ -0,0 +1,161 @@
...
+		nodeWrapper := nodeWrapper.New(&nodeList.Items[0], cs)
+		err = nodeWrapper.SoftRebootNodeAndWaitForDisconnect(ns)
+		Expect(err).ToNot(HaveOccurred())
+
+		nodeWrapper.WaitForNodeReady()
+
+		output, err := firewall.RulesList(cs, nodeList.Items[0])
+		Expect(err).ToNot(HaveOccurred())
+
+		if strings.Contains(string(output), "table inet openshift_filter") &&
```
SchSeba, line 135: this should be a test so use the expect here not a print that can get miss

test/e2e/e2e_tests_suite_test.go
SchSeba: please use the generic client here

test/pkg/firewall/firewall.go
```diff
@@ -0,0 +1,212 @@
...
+	"sync"
+
+	"github.com/openshift-kni/commatrix/pkg/client"
+	"github.com/openshift-kni/commatrix/pkg/consts"
+	"github.com/openshift-kni/commatrix/pkg/utils"
+
+	"github.com/openshift-kni/commatrix/pkg/types"
+
+	v1 "k8s.io/api/core/v1"
+)
```
SchSeba, line 16:
```
	"fmt"
	"os"
	"path/filepath"
	"sync"

	v1 "k8s.io/api/core/v1"

	"github.com/openshift-kni/commatrix/pkg/client"
	"github.com/openshift-kni/commatrix/pkg/consts"
	"github.com/openshift-kni/commatrix/pkg/utils"
	"github.com/openshift-kni/commatrix/pkg/types"
```

test/pkg/firewall/firewall.go
```diff
@@ -0,0 +1,212 @@
...
+
+func ApplyRulesToNode(cs *client.ClientSet, ns string, NFTtable []byte, nodeName string, wg *sync.WaitGroup) error {
+	defer wg.Done()
```
SchSeba, line 19: do not send the wg here just call it in the test.
this function doesn't needs to know about a waitGroup

test/pkg/firewall/firewall.go
```diff
@@ -0,0 +1,212 @@
...
+	"github.com/openshift-kni/commatrix/pkg/client"
+	"github.com/openshift-kni/commatrix/pkg/consts"
+	"github.com/openshift-kni/commatrix/pkg/utils"
+
+	"github.com/openshift-kni/commatrix/pkg/types"
+
+	v1 "k8s.io/api/core/v1"
+)
+
+func ApplyRulesToNode(cs *client.ClientSet, ns string, NFTtable []byte, nodeName string, wg *sync.WaitGroup) error {
```
SchSeba, line 18: please create a struct called firewall and give him the client and the namespace.

to this function only send the nftable and the nodeName to apply

test/pkg/firewall/firewall.go
```diff
@@ -0,0 +1,212 @@
...
+	defer wg.Done()
+
+	utilsHelpers := utils.New(cs)
```
SchSeba, line 21: this also needs to be in the firewall struct so you can have mocks for unit-tests in the future

test/pkg/firewall/firewall.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+
+	defer func() {
+		err := utilsHelpers.DeletePod(debugPod)
+		if err != nil {
+			fmt.Printf("failed cleaning debug pod %s: %v", debugPod, err)
+		}
+	}()
+
+	command := []string{
```
SchSeba, line 35: create the file dirrectly on the host `/host/etc/nftables/firewall.nft` why you need to copy the file after that?

test/pkg/firewall/firewall.go
```diff
@@ -0,0 +1,212 @@
...
+		"bash", "-c",
+		fmt.Sprintf("mkdir -p nft && echo '%s' > nft/firewall.nft", string(NFTtable)),
+	}
+
+	_, err = utilsHelpers.RunCommandOnPod(debugPod, command)
```
SchSeba, line 40: please create an helper function to will do the `/bin/bash -c` and you only send the commands you want

test/pkg/firewall/firewall.go
```diff
@@ -0,0 +1,212 @@
...
+
+	command = []string{
+		"bash", "-c", "/usr/sbin/nft -f nft/firewall.nft",
+	}
+	_, err = utilsHelpers.RunCommandOnPod(debugPod, command)
+	if err != nil {
+		return fmt.Errorf("failed to apply rule set %q on node %s: %w", command, nodeName, err)
+	}
+
+	output, err := nftList(debugPod, utilsHelpers)
```
SchSeba, line 57: this functions needs to be part of the firewall struct object so you only send the debugPod

test/pkg/firewall/firewall.go
```diff
@@ -0,0 +1,212 @@
...
+		"/usr/sbin/nft list ruleset > /etc/nftables.conf",
+	}
+
+	_, err = utilsHelpers.RunCommandOnPod(debugPod, command)
+	if err != nil {
+		return fmt.Errorf("failed to save NFT ruleset to file on node %s: %w", nodeName, err)
+	}
+
+	command = []string{"bash", "-c",
+		"systemctl enable nftables",
```
SchSeba, line 88: please document this one that is not supported in openshift.

you can't just enable a system service this will not survive a cluster upgrade when applying a new RHCOS image we need to find a different solution I think.

or understand how we expect the users to configure the nftables on the machines

SchSeba: cc @yuvalk

test/pkg/firewall/firewall.go
```diff
@@ -0,0 +1,212 @@
...
+
+	err = saveNFTablesRules("commatrix", "nftables-"+nodeName, string(output))
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func RulesList(cs *client.ClientSet, node v1.Node) ([]byte, error) {
```
SchSeba, line 103: this should be part of the firewall struct so you don't need to send the client and also not creating a new utils

test/pkg/firewall/firewall.go
```diff
@@ -0,0 +1,212 @@
...
+		return nil, err
+	}
+
+	return output, nil
+}
+
+func editNftablesConf(debugPod *v1.Pod, u utils.UtilsInterface) error {
+	// Command to add the include statement to nftables.conf
+	command := []string{
+		"sh", "-c", `echo 'include "/etc/nftables/firewall.nft"' >> /host/etc/sysconfig/nftables.conf`,
```
SchSeba, line 160: this is riskly you can't run this twice you need to check if the include already exist I think

SchSeba: also why you do `sh -c` and in all the others you do `/bin/bash -c`

test/pkg/firewall/firewall.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+	_, err := u.RunCommandOnPod(debugPod, command)
+	if err != nil {
+		return fmt.Errorf("failed to edit nftables.conf on debug pod: %w", err)
+	}
+	return nil
+}
+
+func installNFT(debugPod *v1.Pod, u utils.UtilsInterface) error {
+	installCmd := []string{"bash", "-c", "yum update -y && yum install -y nftables"}
```
SchSeba, line 170: we should not install stuff instide the container this will not work on disconnected env.
you can access the host and run the nftables commands from the host (chroot /host nftables)

test/pkg/firewall/firewall.go
```diff
@@ -0,0 +1,212 @@
...
+	filePath := filepath.Join(folderPath, fileName)
+	err = os.WriteFile(filePath, []byte(content), 0644)
+	if err != nil {
+		return fmt.Errorf("failed to save file %s: %w", filePath, err)
+	}
+
+	return nil
+}
+
+func nftList(debugPod *v1.Pod, u utils.UtilsInterface) ([]byte, error) {
```
SchSeba, line 199: this needs to be part of the firewall struct

test/pkg/node/node.go
```diff
@@ -0,0 +1,118 @@
...
+	node.WaitForNodeNotReady()
+	return nil
+}
+
+// WaitForNodeNotReady waits for the node to be in the NotReady state in Kubernetes.
+func (node *NodeWrapper) WaitForNodeNotReady() {
+	log.Printf("Waiting for node %s to be in NotReady state", node.Node.Name)
+
+	timeout := 5 * time.Minute
+	gomega.Eventually(func() bool {
```
SchSeba, line 73: you can't use gomega here as it's under package not test folder.

if you want to have a wait function for the functional tests please create a test/utils/(node/cluster) package and use the eventually function there

aabughosh: its under test/pkg/node
so its still under test

Makefile
```diff
@@ -52,6 +52,10 @@ ginkgo:
 deps-update:
 	go mod tidy
 
+e2e-test:
+	cd test/e2e && ginkgo bootstrap
```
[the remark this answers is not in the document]
aabughosh: i dont want to run it on the 
test:
	GOFLAGS="" go test ./...

test/pkg/firewall/firewall.go
aabughosh: its from here 
https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/9/html/configuring_firewalls_and_packet_filters/getting-started-with-nftables_firewall-packet-filters#including-files-in-nftables-scripts_writing-and-executing-nftables-scripts

test/pkg/firewall/firewall.go
```diff
@@ -0,0 +1,181 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to save rule set to file on node %s: %w", nodeName, err)
+	}
+
+	_, err = firewall.runBashCommandOnPod(debugPod, "nft -f /etc/nftables/firewall.nft", "/host")
+	if err != nil {
+		return fmt.Errorf("failed to apply rule set on node %s: %w", nodeName, err)
+	}
+
+	output, err := firewall.nftList(debugPod)
```
sabinaaledort, line 66: All the firewall.* funcs are internal, why applying them of the struct?

sabinaaledort: It makes the code complicated and less readable instead of just receive ns for example, the code is full of firewall.*

test/pkg/firewall/firewall.go
```diff
@@ -0,0 +1,181 @@
...
+}
+
+func New(ns string, utilsHelpers utils.UtilsInterface) *Firewall {
+	return &Firewall{
+		ns:           ns,
+		utilsHelpers: utilsHelpers,
+	}
+}
+
+func (firewall *Firewall) runBashCommandOnPod(debugPod *v1.Pod, command string, chrootDir string) ([]byte, error) {
```
sabinaaledort, line 27: What is chrootDir?

test/pkg/node/node.go
```diff
@@ -0,0 +1,118 @@
...
+		updatedNode, err := node.cs.Nodes().Get(context.TODO(), node.Node.Name, metav1.GetOptions{})
```
sabinaaledort, line 74: Just an example of why I think this struct is not needed and not used, especially that this is an internal func. This is very confusing and makes the code less readable due to the use of the struct, instead of simply having:
cs.Nodes().Get(context.TODO(), node.Name, metav1.GetOptions{})

test/pkg/firewall/firewall.go
aabughosh: there is commands that we are ruuning on chroot /host

test/pkg/firewall/firewall.go
sabinaaledort: Why is this struct needed? It's not keeping any data related to firewall. All the firewall.* funcs are internal, why applying them of the struct?
It makes the code complicated and less readable instead of just receive ns for example, the code is full of firewall.*
Another example of how the code is less readable this way
https://github.com/openshift-kni/commatrix/pull/36/files#r1726484789

Makefile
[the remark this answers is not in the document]
aabughosh: yes i can but in ur local if u want to run the unit test this will run that reboot test taht maybe u dont want to run

test/e2e/e2e_tests_suite_test.go
```diff
@@ -0,0 +1,167 @@
...
+
+		ns = "commatix-reboot" + nodeRole
+		nodeWrapper := nodeWrapper.New(&nodeList.Items[0], cs)
+		err = nodeWrapper.SoftRebootNodeAndWaitForDisconnect(ns)
+		Expect(err).ToNot(HaveOccurred())
+
+		nodeWrapper.WaitForNodeReady()
+
+		ns = "check-firewall-" + nodeRole
+		firewallcheck := firewall.New(ns, utilsHelpers)
```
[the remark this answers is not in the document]
aabughosh: removed

test/e2e/e2e_tests_suite_test.go
```diff
@@ -0,0 +1,167 @@
...
+		// Wait for all goroutines to finish
+		wg.Wait()
+
+		By("reboot first node")
+		nodeRole, err := types.GetNodeRole(&nodeList.Items[0])
+		Expect(err).ToNot(HaveOccurred())
+
+		ns = "commatix-reboot" + nodeRole
+		nodeWrapper := nodeWrapper.New(&nodeList.Items[0], cs)
+		err = nodeWrapper.SoftRebootNodeAndWaitForDisconnect(ns)
```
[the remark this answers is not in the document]
sabinaaledort: Do you mean the defer that's running in the end of the test?

test/pkg/firewall/firewall.go
sabinaaledort: Always to /host right? So maybe you can just have a bool saying if you need to change to root?

test/e2e/e2e_tests_suite_test.go
```diff
@@ -0,0 +1,167 @@
...
+)
+
+var _ = BeforeSuite(func() {
+	By("generating the commatrix")
+	var err error
+	cs, err = client.New()
+	Expect(err).NotTo(HaveOccurred())
+
+	deployment := types.MNO
+	isSNO, err := isSNOCluster(cs)
```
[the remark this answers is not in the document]
shirmoran: I agree, we also use those function in the origin tests.

test/e2e/e2e_tests_suite_test.go
```diff
@@ -0,0 +1,161 @@
...
+				return nil
+			})
+
+		}
+		// Wait for all goroutines to finish
+		err = g.Wait()
+		Expect(err).ToNot(HaveOccurred())
+
+		By("reboot first node")
+		Expect(err).ToNot(HaveOccurred())
```
shirmoran, line 121: there is no new error to check after checking the previous error in line 118, this line should probably be removed.

test/pkg/node/node.go
```diff
@@ -0,0 +1,118 @@
...
+	"github.com/onsi/gomega"
+
+	"github.com/openshift-kni/commatrix/pkg/client"
+	"github.com/openshift-kni/commatrix/pkg/consts"
+	"github.com/openshift-kni/commatrix/pkg/utils"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type NodeWrapper struct {
```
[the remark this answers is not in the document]
shirmoran: I agree, using the specific node is sufficient since we always use the same client. Also we only test one node...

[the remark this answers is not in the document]
shirmoran: the tested node can just be saved in a variable in my opinion

pkg/utils/utils.go
```diff
@@ -140,6 +140,9 @@ func (u *utils) RunCommandOnPod(pod *corev1.Pod, command []string) ([]byte, erro
 }
 
 func getPodDefinition(node string, namespace string, image string) *corev1.Pod {
+	terminationGracePeriodSeconds := int64(1)
```
sabinaaledort, line 143: This one is used just once so I don't think a var is needed

shirmoran (approved):

test/pkg/utils/utiles.go
```diff
@@ -0,0 +1,28 @@
+package utils
```
sabinaaledort, line 1: typo in file name, why not storing it in the functional package and use it in main as well?

test/pkg/firewall/firewall.go
```diff
@@ -0,0 +1,165 @@
...
+
+	err = saveNFTablesRules("commatrix-after-reboot", "nftables-"+nodeName, string(output))
+	if err != nil {
+		return nil, err
+	}
+
+	return output, nil
+}
+
+func runBashCommandOnPod(debugPod *v1.Pod, command string, chrootDir bool, utilsHelpers utils.UtilsInterface) ([]byte, error) {
```
sabinaaledort, line 96: Why you are not using RunCommandOnPod func in utils?

test/pkg/firewall/firewall.go
```diff
@@ -0,0 +1,165 @@
...
+	output, err := utilsHelpers.RunCommandOnPod(debugPod, []string{"bash", "-c", fullCommand})
+	if err != nil {
+		return nil, fmt.Errorf("failed to execute command '%s' on node %s: %w", command, debugPod.Spec.NodeName, err)
+	}
+	return output, nil
+}
+
+// Make sure that the icnlude of the new nftables is not there.
+// Then if it not there just add that include.
+func editNftablesConf(debugPod *v1.Pod, utilsHelpers utils.UtilsInterface) error {
```
sabinaaledort, line 114: So it's validating something or editing something? I'm not I understand the comment or this functionality and why it's needed

test/pkg/firewall/firewall.go
```diff
@@ -0,0 +1,165 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to edit nftables.conf on debug pod: %w", err)
+	}
+
+	return nil
+}
+
+// For our test save the nftables to see before and after reboot.
+func saveNFTablesRules(folderName, fileName, content string) error {
+	artifactsDir, ok := os.LookupEnv("ARTIFACT_DIR")
```
sabinaaledort, line 133: This should have a default value in the makefile

sabinaaledort: Why do you need a artifactsDir and folderName?

test/pkg/firewall/firewall.go
```diff
@@ -0,0 +1,165 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to save NFT ruleset to file on node %s: %w", nodeName, err)
+	}
+
+	_, err = runBashCommandOnPod(debugPod, "systemctl enable nftables", true, utilsHelpers)
+	if err != nil {
+		return fmt.Errorf("failed to enable nftables on node %s: %w", nodeName, err)
+	}
+
+	err = saveNFTablesRules("commatrix", "nftables-"+nodeName, string(output))
```
sabinaaledort, line 62: I don't understand why you need this function here instead of just printing to file after generating output here https://github.com/openshift-kni/commatrix/pull/36/files#diff-2d8a2bd5eebac03f17584c31940f9f92f6353546ec5b4f33b494045cf94fecdcR41

test/pkg/firewall/firewall.go
```diff
@@ -0,0 +1,165 @@
...
+
+func nftList(debugPod *v1.Pod, utilsHelpers utils.UtilsInterface) ([]byte, error) {
+	command := "nft list ruleset"
+	output, err := runBashCommandOnPod(debugPod, command, true, utilsHelpers)
+	if err != nil {
+		return nil, fmt.Errorf("failed to list NFT ruleset on node %s: %w", debugPod.Spec.NodeName, err)
+	}
+
+	if len(output) == 0 {
+		return nil, fmt.Errorf("no output from 'nft list ruleset' on node %s: ", debugPod.Spec.NodeName)
```
sabinaaledort, line 161: No output which means the there are no rules and that's why it's failing? Confusing error msg

test/pkg/node/node.go
```diff
@@ -0,0 +1,82 @@
...
+
+// SoftRebootNodeAndWaitForDisconnect soft reboots given node and wait for node to be unreachable.
+func SoftRebootNodeAndWaitForDisconnect(debugPod *v1.Pod, cs *client.ClientSet) error {
+	utilsHelpers := utils.New(cs)
+	nodeName := debugPod.Spec.NodeName
+	rebootcmd := []string{"chroot", "/host", "reboot"}
+
+	_, err := utilsHelpers.RunCommandOnPod(debugPod, rebootcmd)
+	if err != nil {
+		return fmt.Errorf("failed to install nftables on  debugpod on node%s: %w", nodeName, err)
```
sabinaaledort, line 25: wrong err msg

test/pkg/node/node.go
```diff
@@ -0,0 +1,82 @@
...
+	return nil
+}
+
+// WaitForNodeNotReady waits for the node to be in the NotReady state in Kubernetes.
+func WaitForNodeNotReady(nodeName string, cs *client.ClientSet) {
+	log.Printf("Waiting for node %s to be in NotReady state", nodeName)
+
+	timeout := 5 * time.Minute
+	gomega.Eventually(func() bool {
+		updatedNode, err := cs.CoreV1Interface.Nodes().Get(context.TODO(), nodeName, metav1.GetOptions{})
```
sabinaaledort, line 38: Can just be node instead of updatedNode since we don't update anything here

test/pkg/firewall/firewall.go
aabughosh: it will be so much lines

test/pkg/firewall/firewall.go
```diff
@@ -0,0 +1,165 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return output, nil
+}
+
+func runBashCommandOnPod(debugPod *v1.Pod, command string, chrootDir bool, utilsHelpers utils.UtilsInterface) ([]byte, error) {
+	var fullCommand string
+	if chrootDir {
```
sabinaaledort, line 98: Can you hold the chroot /host prefix in a const and add it as a prefix when needed? I think the func name is confusing, RunCommandOnPod also running bash commands, just not as root. You can change the name and update the if but I think it's better to just use a prefix

sabinaaledort, line 98: Can just be:
if chrootDir {
command := fmt.Sprintf("chroot %s /bin/bash -c '%s'", "/host", command)
}

test/e2e/e2e_tests_suite_test.go
```diff
@@ -0,0 +1,146 @@
...
+	cs           *client.ClientSet
+	matrix       *types.ComMatrix
+	isSNO        bool
+	utilsHelpers utils.UtilsInterface
+	nodeList     *corev1.NodeList
+)
+
+var _ = BeforeSuite(func() {
+	By("generating the commatrix")
+	var err error
```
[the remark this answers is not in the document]
aabughosh: that will creat a new cs and not the cs that i put on the vars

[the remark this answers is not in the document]
sabinaaledort: right sorry

pkg/client/client.go
```diff
@@ -1,6 +1,7 @@
 package client
 
 import (
+	clientconfigv1 "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
```
SchSeba, line 4: why you don't just add the openshift API to the scheme?

https://github.com/openshift/api/blob/master/config/install.go#L18

test/e2e/e2e_tests_suite_test.go
```diff
@@ -0,0 +1,156 @@
+package e2e
```
SchSeba, line 1: please check the file name it's a weird name e2e_tests_suite_test.go

test/e2e/e2e_tests_suite_test.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+
+	epExporter, err := endpointslices.New(cs)
+	Expect(err).ToNot(HaveOccurred())
+
+	By("Create commMatrix")
+	commMatrix, err := commatrixcreator.New(epExporter, "", "", infra, deployment)
+	Expect(err).NotTo(HaveOccurred())
+
+	matrix, err = commMatrix.CreateEndpointMatrix()
```
SchSeba, line 75: This is part of the test please move it inside the test section

test/e2e/e2e_tests_suite_test.go
```diff
@@ -0,0 +1,156 @@
...
+				return nil
+			})
+		}
+
+		// Wait for all goroutines to finish
+		err = g.Wait()
+		Expect(err).ToNot(HaveOccurred())
+
+		By("reboot first node")
+
```
SchSeba, line 131: remove new line

test/pkg/firewall/firewall.go
```diff
@@ -0,0 +1,140 @@
...
+	return output, nil
+}
+
+// Checks if the include statement for the new nftables configuration,
+// Is present in the nftables.conf file. If the include statement is not present,
+// Add it the statement to the file.
+func editNftablesConf(debugPod *v1.Pod, utilsHelpers utils.UtilsInterface) error {
+	checkCommand := `grep -q 'include "/etc/nftables/firewall.nft"' /host/etc/sysconfig/nftables.conf`
+	_, err := RunCommandInPod(debugPod, checkCommand, false, utilsHelpers)
+	if err == nil {
```
SchSeba, line 114: This is not right. the `RunCommandInPod` command can failed because of other issues like connection to pod, command run issue etc..

you can have a case where running the exec failed (but the file exist) and you will add it again.

you can do something like `grep 'include "/etc/nftables/firewall.nft"' /host/etc/sysconfig/nftables.conf | wc -l` and check the output is 0 or 1

test/pkg/node/node.go
```diff
@@ -0,0 +1,82 @@
...
+	return nil
+}
+
+// WaitForNodeNotReady waits for the node to be in the NotReady state in Kubernetes.
+func WaitForNodeNotReady(nodeName string, cs *client.ClientSet) {
+	log.Printf("Waiting for node %s to be in NotReady state", nodeName)
+
+	timeout := 5 * time.Minute
+	gomega.Eventually(func() bool {
+		node, err := cs.CoreV1Interface.Nodes().Get(context.TODO(), nodeName, metav1.GetOptions{})
```
SchSeba, line 38: please use the generic client

test/pkg/node/node.go
```diff
@@ -0,0 +1,82 @@
...
+	log.Printf("Node %s is NotReady\n", nodeName)
+}
+
+// WaitForNodeReady waits for the node to be in the Ready state in Kubernetes.
+func WaitForNodeReady(nodeName string, cs *client.ClientSet) {
+	log.Printf("Waiting for node %s to be in Ready state", nodeName)
+
+	timeout := 15 * time.Minute
+	gomega.Eventually(func() bool {
+		node, err := cs.CoreV1Interface.Nodes().Get(context.TODO(), nodeName, metav1.GetOptions{})
```
SchSeba, line 64: please use the generic client

SchSeba (commented):
great work!

I left some small comments

test/e2e/e2e_tests_suite_test.go
aabughosh: we agreed to put it here maybe will have more test in the future  that will use it

pkg/utils/utils.go
```diff
@@ -169,19 +176,30 @@ func getPodDefinition(node string, namespace string, image string) *corev1.Pod {
...
+					TolerationSeconds: ptr.To(tolerationSeconds),
+				},
+			},
+			Volumes: []corev1.Volume{
+				{
+					Name: "host-root",
+					VolumeSource: corev1.VolumeSource{
+						HostPath: &corev1.HostPathVolumeSource{
+							Path: "/",
+							Type: ptr.To(corev1.HostPathType("Directory")), // Ensure the path is a directory
```
bnshr, line 200: You could use some constants for "Directory" here.

pkg/utils/utils.go
```diff
@@ -200,3 +218,23 @@ func getNamespaceDefinition(namespace string) *corev1.Namespace {
 		},
 	}
 }
+
+func (u *utils) IsSNOCluster() (bool, error) {
+	infra := &configv1.Infrastructure{}
+	err := u.Get(context.Background(), clientOptions.ObjectKey{Name: "cluster"}, infra)
```
bnshr, line 224: Same here. Use some constant instead of hardcoded "cluster"

test/e2e/commatrix_suite_test.go
```diff
@@ -0,0 +1,155 @@
...
+
+		g := new(errgroup.Group)
+
+		for _, node := range nodeList.Items {
+			nodeName := node.Name
+			nodeRole, err := types.GetNodeRole(&node)
+			Expect(err).ToNot(HaveOccurred())
+			g.Go(func() error {
+				nftTable := masterNFT
+				if nodeRole == "worker" {
```
bnshr, line 114: Use some constant like WorkerNodeRoleStr

test/pkg/node/node.go
```diff
@@ -0,0 +1,82 @@
...
+	)
+
+	log.Printf("Node %s is NotReady\n", nodeName)
+}
+
+// WaitForNodeReady waits for the node to be in the Ready state in Kubernetes.
+func WaitForNodeReady(nodeName string, cs *client.ClientSet) {
+	log.Printf("Waiting for node %s to be in Ready state", nodeName)
+
+	timeout := 15 * time.Minute
```
bnshr, line 62: Timeouts that are used can declared as constant , for 3*time.Second and 15 * time.Minute

bnshr (changes requested):

test/e2e/commatrix_suite_test.go
```diff
@@ -0,0 +1,155 @@
...
+var _ = AfterSuite(func() {
+	By("Delete Namespace")
+	err := utilsHelpers.DeleteNamespace(consts.DefaultDebugNamespace)
+	Expect(err).ToNot(HaveOccurred())
+})
+
+var _ = Describe("commatrix", func() {
+	It("should apply firewall by blocking all ports except the ones OCP is listening on", func() {
+		By("apply firewall on nodes")
+		masterMat, workerMat := matrix.SeparateMatrixByRole()
```
edcdavid, line 96: Add a description for each step and a By() statement. Also in Beforesuite.
Start the By() text with a verb ending in "ing". For instance By("applying firewall configuration on all nodes")

pkg/utils/utils.go
aabughosh: i thought thats fine to be here im using it just here so maybe we can keep

test/pkg/firewall/firewall.go
```diff
@@ -0,0 +1,152 @@
...
+
+	err = utilsHelpers.WriteFile(filepath.Join(artifactsDir, "nftables-after-reboot-"+nodeName), output)
+	if err != nil {
+		return nil, err
+	}
+
+	return output, nil
+}
+
+func RunCommandInPod(debugPod *v1.Pod, command string, chrootDir bool, utilsHelpers utils.UtilsInterface) ([]byte, error) {
```
sabinaaledort, line 97: `RunCommandInPod` func calling `RunCommandOnPod` func... the naming doesn't explain why it's done like this. `RunRootCommandOnPod` maybe?

test/e2e/commatrix_suite_test.go
```diff
@@ -0,0 +1,162 @@
...
+})
+
+var _ = AfterSuite(func() {
+	By("Deleting Namespace")
+	err := utilsHelpers.DeleteNamespace(consts.DefaultDebugNamespace)
+	Expect(err).ToNot(HaveOccurred())
+})
+
+var _ = Describe("commatrix", func() {
+	It("should apply firewall by blocking all ports except the ones OCP is listening on", func() {
```
sabinaaledort, line 100: `ApplyRulesToNode`, `SoftRebootNodeAndWaitForDisconnect` and `RulesList` funcs are all creating pods and waiting for them to be ready, it's time consuming, please consider reusing the pods

test/pkg/firewall/firewall.go
```diff
@@ -0,0 +1,152 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to save NFT ruleset to file on node %s: %w", nodeName, err)
+	}
+
+	_, err = RunCommandInPod(debugPod, "systemctl enable nftables", true, utilsHelpers)
+	if err != nil {
+		return fmt.Errorf("failed to enable nftables on node %s: %w", nodeName, err)
+	}
+
+	err = utilsHelpers.WriteFile(filepath.Join(artifactsDir, "nftables-"+nodeName), output)
```
sabinaaledort, line 63: I think the output can be written to the log just after it's created, if a step fails in between it won't be written and we will get less debug logs

test/pkg/firewall/firewall.go
```diff
@@ -0,0 +1,152 @@
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func RulesList(nodeName, artifactsDir string, utilsHelpers utils.UtilsInterface) ([]byte, error) {
```
sabinaaledort, line 71: `RulesList` is a general name to a function that can only be called after reboot since you set the that in the file name, please consider using making this func and `nftList` a single func that can be reused

test/e2e/commatrix_suite_test.go
```diff
@@ -0,0 +1,162 @@
...
+	workerNodeRole = "worker"
+	tableName      = "table inet openshift_filter"
+	chainName      = "chain OPENSHIFT"
+)
+
+var _ = BeforeSuite(func() {
+	By("create output folder")
+	artifactsDir = os.Getenv("ARTIFACT_DIR")
+	if artifactsDir == "" {
+		log.Println("env var ARTIFACT_DIR is not set, using default value")
```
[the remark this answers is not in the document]
aabughosh: artifactsDir = ""
that will create on the place that the e2e is

PR #37, opened 2024-08-13
Initialize a default logger if needed

aabughosh:


pkg/utils/logutils.go
```diff
@@ -25,5 +25,10 @@ func InitializeLog(isDebug bool) {
 
 // Logger returns the configured Logrus logger instance.
 func Logger() *logrus.Logger {
+	if logger == nil {
```
sabinaaledort, line 28: I don't think it's a clean solution, why we have an init for this object if we do it here anyway?
If that's an object that has an init all the other objects using it should hold an instance and call the init. 
Worth thinking about removing the logs from the code and keeping a single logger in main to print steps/errors.

aabughosh: What is the point with that?
We want to it to print the debugs one on debug mode not sure about the removing

sabinaaledort: We are working with objects, so every object using logger should create an instance of logger, logger initing itself is not the correct way to work with objects.

pkg/commatrix-creator/commatrix.go
```diff
@@ -41,6 +41,9 @@ func New(exporter *endpointslices.EndpointSlicesExporter, customEntriesPath stri
 // port number, namespace, service name, pod, container, node role, and flow optionality for OpenShift.
 func (cm *CommunicationMatrixCreator) CreateEndpointMatrix() (*types.ComMatrix, error) {
 	log := utils.Logger() // Get the logger instance from utils
+	if log == nil {
```
sabinaaledort, line 44: Why do you check for nil? If a object using a logger it should init and hold a logger object

PR #38, opened 2024-08-18
Add matrix-diff package

sabinaaledort:
Adding a matrix-diff struct and functions to simplify the use of matrix diff in tests and other scenarios.

### DIFF
--- a/pkg/matrix-diff/matrix-diff.go
+++ b/pkg/matrix-diff/matrix-diff.go
@@ -1,0 +1,74 @@
+package matrixdiff
+
+import (
+	"fmt"
+
+	"github.com/openshift-kni/commatrix/pkg/types"
+)
+
+type status int
+
+// Both = cd's present in both mat1 and mat2.
+// UniqueA = cd's present in mat1 but not in mat2.
+// UniqueB = cd's present in mat2 but not in mat1.
+const (
+	both status = iota
+	uniqueA
+	uniqueB
+)
+
+type MatrixDiff struct {
+	types.ComMatrix
+	cdToStatus map[string]status
+}
+
+// Generates the diff between mat1 to mat2.
+func Generate(mat1 *types.ComMatrix, mat2 *types.ComMatrix) MatrixDiff {
+	matrix := types.ComMatrix{}
+	epsStatus := map[string]status{}
+
+	for _, cd := range mat1.Matrix {
+		matrix.Matrix = append(matrix.Matrix, cd)
+		epsStatus[cd.String()] = both
+
+		if !mat2.Contains(cd) {
+			epsStatus[cd.String()] = uniqueA
+		}
+	}
+
+	for _, cd := range mat2.Matrix {
+		matrix.Matrix = append(matrix.Matrix, cd)
+		epsStatus[cd.String()] = both
+
+		if !mat1.Contains(cd) {
+			epsStatus[cd.String()] = uniqueB
+		}
+	}
+
+	matrix.SortAndRemoveDuplicates()
+
+	return MatrixDiff{matrix, epsStatus}
+}
+
+func (m *MatrixDiff) String() (string, error) {
+	colNames, err := types.GetComMatrixHeadersByFormat(types.FormatCSV)
+	if err != nil {
+		return "", fmt.Errorf("error getting commatrix CSV tags: %v", err)
+	}
+	diff := colNames + "\n"
+
+	for _, cd := range m.Matrix {
+		switch m.cdToStatus[cd.String()] {
+		case both:
+			diff += fmt.Sprintf("%s\n", cd)
+		case uniqueA:
+			// add "+" before cd's present in mat1 but not in mat2.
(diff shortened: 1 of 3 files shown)


pkg/matrix-diff/matrix-diff.go
```diff
@@ -0,0 +1,75 @@
...
+	missing
+)
+
+type MatrixDiff struct {
+	matrix        []types.ComDetails
+	cdToEPSStatus map[string]epsStatus
+}
+
+// Generates the diff between mat1 to mat2.
+func Generate(mat1 *types.ComMatrix, mat2 *types.ComMatrix) (MatrixDiff, error) {
```
aabughosh, line 27: its not better to have  func (m *ComMatrix) Generate(other *ComMatrix) (MatrixDiff, error){}?
like from one matrix we are generating diff from the other on that format its more clear

pkg/matrix-diff/matrix-diff.go
```diff
@@ -0,0 +1,75 @@
...
+		return "", fmt.Errorf("error getting commatrix CSV tags: %v", err)
+	}
+	diff := colNames + "\n"
+
+	for _, cd := range m.matrix {
+		switch m.cdToEPSStatus[cd.String()] {
+		case used:
+			diff += fmt.Sprintf("%s\n", cd)
+		case notUsed:
+			// add "+" before cds that have an endpointslice but the port is not open on the node
```
aabughosh, line 66: its need to be general not diff from endpointslice?
what if we have 2 martix and we want to show diff maybe from 2 different  cluster?

pkg/matrix-diff/matrix-diff.go
sabinaaledort: To do that you have to keep this function in the types files (where ComMatrix is defined) which is too random and general (and should be gone soon) so I prefer creating a new package for diff

pkg/matrix-diff/matrix-diff.go
sabinaaledort: Matrix represents endpointslices in the cluster, what do you mean by general?

aabughosh: like diff matrix is always between endpointslice matrix?

pkg/types/types.go
```diff
@@ -329,32 +264,33 @@ func (m *ComMatrix) ToNFTables() ([]byte, error) {
 	return []byte(result), nil
 }
 
-// SortAndRemoveDuplicates removes duplicates in the matrix and sort it.
-func (m *ComMatrix) SortAndRemoveDuplicates() {
+// SortAndRemoveDuplicates sorts ComDetails and removes duplicates.
+func SortAndRemoveDuplicates(cds []ComDetails) []ComDetails {
```
aabughosh, line 268: why u changed this (m *ComMatrix) SortAndRemoveDuplicates()?

sabinaaledort: To use it in the new matrix diff package

pkg/matrix-diff/matrix-diff.go
```diff
@@ -0,0 +1,75 @@
...
+// NotUsed = the port is not open on the node.
+// Missing = no endpointslice defined for an open port.
+const (
+	used epsStatus = iota
+	notUsed
+	missing
+)
+
+type MatrixDiff struct {
+	matrix        []types.ComDetails
```
aabughosh, line 22: why that is not ComMatrix? like that will be the same

pkg/types/types.go
aabughosh: yah but u can create on the diff as ComMatrix and not []ComDetails

pkg/matrix-diff/matrix-diff.go
shirmoran: We also do diff between the documented matrix and the endpoint slice matrix, so maybe the comment is not accurate.

pkg/matrix-diff/matrix-diff.go
shirmoran, line 22: a `[]types.ComDetails` is basically a `comMatrix`, so I maybe it's will be better saving it that way, and by that allowing to use it's methods instead of changing the `SortAndRemoveDuplicates` method to a function for example?

pkg/matrix-diff/matrix-diff.go
```diff
@@ -0,0 +1,75 @@
...
+// Used = the port is open on the node.
+// NotUsed = the port is not open on the node.
+// Missing = no endpointslice defined for an open port.
+const (
+	used epsStatus = iota
+	notUsed
+	missing
+)
+
+type MatrixDiff struct {
```
shirmoran, line 21: a []types.ComDetails is basically a comMatrix, so I maybe it's will be better saving it that way, and by that allowing to use it's methods instead of changing the SortAndRemoveDuplicates method to a function for example?

aabughosh: agree with shir

oribon (commented):
small nit, other than that lgtm

PR #41, opened 2024-08-21
Add a test for the ssGenrate function on listeningsockets

aabughosh:


shirmoran (approved):

pkg/listening-sockets/listening-sockets_test.go
```diff
@@ -0,0 +1,166 @@
...
+								"io.kubernetes.pod.namespace": "test-namespace"
+							}
+						}
+					]
+				}`), nil
+			}
+			if strings.HasPrefix(cmd[2], "cat /proc/") && strings.Contains(cmd[2], "/cgroup") {
+				return output, nil
+			}
+			if cmd[2] == "ss -anpltH" {
```
sabinaaledort, line 88: You can avoid these ifs by replacing these commands like in the next expects

pkg/listening-sockets/listening-sockets_test.go
```diff
@@ -0,0 +1,166 @@
...
+		Client:          fakeClient,
+		CoreV1Interface: fakeClientset.CoreV1(),
+	}
+
+	ctrlTest := gomock.NewController(t)
+	defer ctrlTest.Finish()
+
+	mockUtils := mock_utils.NewMockUtilsInterface(ctrlTest)
+
+	tcpOutput := []byte(`LISTEN 0      4096      127.0.0.1:8797  0.0.0.0:* users:(("machine-config-",pid=3534,fd=3))
```
sabinaaledort, line 58: Are these consts? Maybe better use a DescribeTable

pkg/listening-sockets/listening-sockets_test.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	mockUtils.EXPECT().
+		CreatePodOnNode(gomock.Any(), consts.DefaultDebugNamespace, consts.DefaultDebugPodImage).
+		Return(mockPod, nil).AnyTimes()
+
+	mockUtils.EXPECT().DeletePod(mockPod).Return(nil).AnyTimes()
+
+	connectionCheck, err := NewCheck(clientset, mockUtils, "/some/dest/dir")
+	assert.NoError(t, err)
```
sabinaaledort, line 122: let's use ginkgo for the unit tests as well

pkg/listening-sockets/listening-sockets_test.go
```diff
@@ -0,0 +1,166 @@
...
+
+	ssMat, ssOutTCP, ssOutUDP, err := connectionCheck.GenerateSS()
+
+	expectedTCPOutput := `node: test-node
```
sabinaaledort, line 126: Same, should be consts or better in a DescribeTable

pkg/listening-sockets/listening-sockets_test.go
```diff
@@ -0,0 +1,166 @@
+package listeningsockets
```
sabinaaledort, line 1: Can you fix the file name?

pkg/listening-sockets/listening-sockets_test.go
aabughosh: Because  its from same function that we want to mock maybe better to have it like this?

sabinaaledort: If you can have them in a one liner instead of 5 ifs I think it makes the code more readable

pkg/listening-sockets/listening-sockets_test.go
```diff
@@ -0,0 +1,205 @@
...
+		Entry("Case 1", ssTestCase{
+			name:              "Case 1",
+			expectedTCPOutput: expectedTCPOutput,
+			expectedUDPOutput: expectedUDPOutput,
+			expectedssMat:     expectedssMat,
+		}),
+	)
+
+	AfterEach(func() {
+		ctrlTest.Finish()
```
sabinaaledort, line 195: AfterEach should usually be at the top next to BeforeEach for better readability if we add more tests in between

pkg/listening-sockets/listening-sockets_test.go
```diff
@@ -0,0 +1,205 @@
...
+						return []byte(crictlExecCommandOut), nil // container data output
+
+					case strings.HasPrefix(cmd[2], "cat /proc/") &&
+						strings.Contains(cmd[2], "/cgroup"):
+						return []byte(procExecCommandOutput), nil // pid data output
+
+					case cmd[2] == "ss -anpltH":
+						return []byte(tcpExecCommandOutput), nil // tcp output
+
+					case cmd[2] == "ss -anpluH":
```
sabinaaledort, line 151: Is it possible to make all these as the others?
mockUtils.EXPECT().
		RunCommandOnPod("ss -anpltH").
		Return([]byte(tcpExecCommandOutput), nil).
		AnyTimes()

pkg/listening-sockets/listening-sockets_test.go
```diff
@@ -0,0 +1,205 @@
...
+	})
+
+	DescribeTable("should generate the correct ss tcp, udp output and the correct ssMatrix",
+		func(tc ssTestCase) {
+
+			// RunCommandOnPod had more that one calling and in each call we want other output
+			mockUtils.EXPECT().RunCommandOnPod(gomock.Any(), gomock.Any()).DoAndReturn(
+				func(pod *v1.Pod, cmd []string) ([]byte, error) {
+					switch {
+					case len(cmd) > 2 && strings.HasPrefix(cmd[2], "crictl ps -o json --id"):
```
sabinaaledort, line 141: This is not clear to me

pkg/listening-sockets/listening-sockets_test.go
```diff
@@ -0,0 +1,205 @@
...
+const (
+	tcpExecCommandOutput = (`LISTEN 0      4096      127.0.0.1:8797  0.0.0.0:* users:(("machine-config-",pid=3534,fd=3))                
+	LISTEN 0      4096      127.0.0.1:8798  0.0.0.0:* users:(("machine-config-",pid=3534,fd=13))               
+	LISTEN 0      4096      127.0.0.1:9100  0.0.0.0:* users:(("node_exporter",pid=4147,fd=3))`)
+
+	udpExecCommandOutput = (`UNCONN 0      0           0.0.0.0:111   0.0.0.0:* users:(("rpcbind",pid=1399,fd=5),("systemd",pid=1,fd=78))
+	UNCONN 0      0         127.0.0.1:323   0.0.0.0:* users:(("chronyd",pid=1015,fd=5))                        
+	UNCONN 0      0      10.46.97.104:500   0.0.0.0:* users:(("pluto",pid=2115,fd=21))`)
+
+	procExecCommandOutput = (`1: /system.slice/crio-1234567890abcdef.scope
```
sabinaaledort, line 39: Is that something we need to validate?

pkg/listening-sockets/listening-sockets_test.go
```diff
@@ -0,0 +1,205 @@
...
+
+			ssMat, ssOutTCP, ssOutUDP, err := connectionCheck.GenerateSS()
+			Expect(err).NotTo(HaveOccurred())
+
+			Expect(normalizeOutput(string(ssOutTCP))).To(Equal(normalizeOutput(tc.expectedTCPOutput)))
+			Expect(normalizeOutput(string(ssOutUDP))).To(Equal(normalizeOutput(tc.expectedUDPOutput)))
+			Expect(ssMat.Matrix).To(Equal(tc.expectedssMat))
+		},
+
+		Entry("Case 1", ssTestCase{
```
sabinaaledort, line 186: I really like this format with ginkgo, it is much easier to read the test now. I think I was wrong with the describe table, don't think we have more tests to add here, so can just be an It

pkg/listening-sockets/listening-sockets_test.go
```diff
@@ -0,0 +1,205 @@
...
+	ctrlTest  *gomock.Controller
+
+	mockPod = &v1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "mock-pod",
+			Namespace: "mock-namespace",
+		},
+	}
+
+	expectedssMat = []types.ComDetails{
```
sabinaaledort, line 79: expectedSSMat

pkg/listening-sockets/listening-sockets_test.go
aabughosh: yah i tried that it will go always to one of them
we are creating fake output from same function so every time im checking what is the cmd that we want to fake

pkg/listening-sockets/listening-sockets_test.go
aabughosh: no this is how the output need to be its need to be with crio and a number this is why i created that on that way
this is the output of the cmd command

pkg/listening-sockets/listening-sockets_test.go
aabughosh: on toComDeatails function there a calling to get container data one of the data that we are getting is this
	cmd := fmt.Sprintf("crictl ps -o json --id %s", containerID)
and output should be on that format 
type ContainerInfo struct {
	Containers []struct {
		Labels struct {
			ContainerName string `json:"io.kubernetes.container.name"`
			PodName       string `json:"io.kubernetes.pod.name"`
			PodNamespace  string `json:"io.kubernetes.pod.namespace"`
		} `json:"labels"`
	} `json:"containers"`
}

pkg/listening-sockets/listening-sockets_test.go
aabughosh: cmd := fmt.Sprintf("cat /proc/%s/cgroup", pid)
output need to have like this
	re := regexp.MustCompile(`crio-([0-9a-fA-F]+)\.scope`)

SchSeba (commented):
from my side LGTM

PR #46, opened 2024-09-08
Updated modules to match origin repo

shirmoran:
This PR updates go modules to match the modules in the `openshift/origin` repo - to avoid mismatch versions that results conflicts in running the different tests.


aabughosh (approved):
/approve
/lgtm

PR #50, opened 2024-09-17
Add doc comparison to generating commatrix origin test

shirmoran:
This PR adds the origin test which compares between the documented communication matrix to the generated one (see [origin repo PR](https://github.com/openshift/origin/pull/28946)) , to our e2e tests.

Also, the different tests are dived to two suite in different files:

1. validation_test.go: The "Validation" suite contains the generated communication matrix comparison tests (to both ss  and documented commatrices).
2. nftables_test.go: The "Nftables" contains the nftables test (applying firewall by blocking on all ports except the ones OCP is listening on)

To run the tests use:

`make e2e-test SUITE="suite-name"`

with the following "suite-name" options corresponding to the wanted suite\s: "all", "Validation", "Nftables"


test/e2e/commatrix_suite_test.go
```diff
@@ -215,6 +344,34 @@ var _ = Describe("commatrix", func() {
 	})
 })
 
+// getClusterVersion return cluster's Y stream version.
+func getClusterVersion(cs *client.ClientSet) (string, error) {
```
aabughosh, line 348: can u put this in utils?
i will use it on my task also
i think ur task will be merge before mine so can u do that?

test/e2e/commatrix_suite_test.go
```diff
@@ -108,8 +152,93 @@ var _ = AfterSuite(func() {
...
+			docComDetailsList = excludeStaticEntriesWithGivenNodeRole(docComDetailsList, &types.ComMatrix{Matrix: types.CloudStaticEntriesWorker}, "worker")
+			docComDetailsList = excludeStaticEntriesWithGivenNodeRole(docComDetailsList, &types.ComMatrix{Matrix: types.CloudStaticEntriesMaster}, "master")
+		} else {
+			docComDetailsList = excludeStaticEntriesWithGivenNodeRole(docComDetailsList, &types.ComMatrix{Matrix: types.BaremetalStaticEntriesWorker}, "worker")
+			docComDetailsList = excludeStaticEntriesWithGivenNodeRole(docComDetailsList, &types.ComMatrix{Matrix: types.BaremetalStaticEntriesMaster}, "master")
+		}
+		docComMatrix := &types.ComMatrix{Matrix: docComDetailsList}
+
+		By("generating diff between matrices for testing purposes")
+		diff := matrixdiff.Generate(commatrix, docComMatrix)
```
sabinaaledort, line 219: Maybe to make it more readable worth changing the order here, so docs is first, to align with the name of the file next and to generate UniquePrimary first

test/e2e/commatrix_suite_test.go
```diff
@@ -108,8 +152,93 @@ var _ = AfterSuite(func() {
...
+		diffStr, err := diff.String()
+		Expect(err).ToNot(HaveOccurred())
+		err = os.WriteFile(filepath.Join(artifactsDir, "doc-diff-new"), []byte(diffFileComments+diffStr), 0644)
+		Expect(err).ToNot(HaveOccurred())
+
+		By("comparing new and documented commatrices")
+		// Get ports that are in the documented commatrix but not in the generated commatrix.
+		notUsedPortsMat := diff.GenerateUniqueSecondary()
+		if len(notUsedPortsMat.Matrix) > 0 {
+			logrus.Warningf("the following ports are documented but are not used: \n %s", notUsedPortsMat)
```
sabinaaledort, line 229: Can you remove the space between \n and %s? There is an extra space in the first line when it's printed

test/e2e/commatrix_suite_test.go
```diff
@@ -108,8 +152,93 @@ var _ = AfterSuite(func() {
...
+		}
+
+		// Get ports that are in the generated commatrix but not in the documented commatrix.
+		missingPortsMat := diff.GenerateUniquePrimary()
+		if len(missingPortsMat.Matrix) > 0 {
+			err := fmt.Errorf("the following ports are used but are not documented: \n %s", missingPortsMat)
```
sabinaaledort, line 235: Same

test/e2e/commatrix_suite_test.go
shirmoran: but as we discussed we want to add `+` where we need to add to the doc, and `-` where we need to remove from the doc, so for that the generated commatrix should be the primary.

PR #60, opened 2024-10-29
Nftables E2E test: allow adding custom ports/ranges to firewall

aabughosh:


test/e2e/nftables_test.go
```diff
@@ -16,22 +18,43 @@ import (
 )
 
 var (
-	workerNodeRole = "worker"
-	tableName      = "table inet openshift_filter"
-	chainName      = "chain OPENSHIFT"
+	workerNodeRole    = "worker"
+	tableName         = "table inet openshift_filter"
```
sabinaaledort, line 22: Should we keep those vars in the code consts?

sabinaaledort: Ah sorry I missed that, thought it's for adding

cgoncalves:
What is the requirement/use case that may require adding extra rules to nftables?

aabughosh:
> What is the requirement/use case that may require adding extra rules to nftables?

On the origin test there was few ports that needed to be add for passing the tests need to allow them

sabinaaledort:
> Which ports? Are those ports associated with cluster services? If yes, why isn't the communication matrix tool detecting and outputting them? I'm just trying to understand if there is a gap in the tool in the first place.

The idea is to allow adding more ports to the list for additional testing by the user. For example in our case when running origin tests some ports from the dynamic port ranges are used. These ranges are also documented here https://docs.openshift.com/container-platform/4.16/installing/install_config/configuring-firewall.html#network-flow-matrix_configuring-firewall

cgoncalves:
Thanks for explaining the requirement. It helps focus the review on what is important.
I wanted to ensure the communication tool was not leaving behind ports required by cluster services. LGTM.

test/e2e/nftables_test.go
```diff
@@ -88,3 +100,39 @@ var _ = Describe("Nftables", func() {
...
+
+func AddPortsToNFTables(nftables []byte, extraNFTablesFile string) ([]byte, error) {
+	nftStr := string(nftables)
+
+	insertPoint := "# Logging and default drop"
+	if !strings.Contains(nftStr, insertPoint) {
+		return nftables, fmt.Errorf("insert point not found in nftables configuration")
+	}
+
+	/*
```
sabinaaledort, line 112: When I talked about an example I meant something like the samples/custom-entries files we have, can be a file or somewhere in the top part of the test file

PR #65, opened 2024-11-20
create ocp version to be const

aabughosh:


test/pkg/firewall/firewall.go
```diff
@@ -13,6 +13,8 @@ import (
 	v1 "k8s.io/api/core/v1"
 )
 
+const ocpVersion = "4.16.0"
```
sabinaaledort, line 16: This is the butane spec version, let's call this var accordingly

test/pkg/firewall/firewall.go
```diff
@@ -47,7 +49,7 @@ func NftListAndWriteToFile(debugPod *v1.Pod, utilsHelpers utils.UtilsInterface,
 
 func CreateMachineConfig(c *client.ClientSet, NFTtable []byte, artifactsDir, nodeRolde, clusterVersion string,
```
sabinaaledort, line 50: clusterVersion can be removed

sabinaaledort (commented):
You can also update ValidateClusterVersionAndMachineConfiguration func, no need to return the clusterVersion with this change

PR #71, opened 2024-12-08
Temp for logs

aabughosh:


test/pkg/firewall/firewall.go
```diff
@@ -28,23 +32,60 @@ func RunRootCommandOnPod(debugPod *v1.Pod, command string, chrootDir bool, utils
...
+
+		if !strings.Contains(string(output), tableName) || !strings.Contains(string(output), chainName) {
+			log.Printf("Required table %s or chain %s not found in nftables output.", tableName, chainName)
+			return false
+		}
+
+		return true
+	}, 2*time.Minute, 10*time.Second).Should(gomega.BeTrue(), "Timeout while waiting for nftables to contain table %s and chain %s", tableName, chainName)
+
+	if cmdErr != nil {
```
sabinaaledort, line 60: This is checked in the Eventually, if it fails there it will timeout and fail and won't get here

test/pkg/firewall/firewall.go
```diff
@@ -28,23 +32,60 @@ func RunRootCommandOnPod(debugPod *v1.Pod, command string, chrootDir bool, utils
...
+		return false, fmt.Errorf("failed to list nft ruleset: %w", cmdErr)
 	}
 
-	if len(output) == 0 {
-		return nil, fmt.Errorf("no nft rules on node %s: ", debugPod.Spec.NodeName)
+	if !strings.Contains(string(output), tableName) || !strings.Contains(string(output), chainName) {
```
sabinaaledort, line 64: Same

PR #81, opened 2024-12-17
Allow adding extra endpoint slices for open ports in e2e test

shirmoran:
This PR allow specifying an extra `comdetails` file, that contains open ports which does not have endpoint.
Those ports will be ignored when comparing the differences between the open ports matrix with the endpoint slices communication matrix, to avoid unnecessary failures. 
Also, those ports will be added to the nftable rules in the e2e nftables test.
This option is needed for certain ports only.

The extra ports should be stored in a file (csv\json\yaml), and its path and format should be passed as the following env vars when running the e2e-test make target:
`OPEN_PORTS_TO_IGNORE_FILE`, `OPEN_PORTS_TO_IGNORE_FORMAT`

Additionally, this PR moves the extra nftables env vars extraction to the test's file.

### DIFF
--- a/test/e2e/e2e_suite_test.go
+++ b/test/e2e/e2e_suite_test.go
@@ -17,16 +17,16 @@ import (
 )
 
 var (
-	cs                      *client.ClientSet
-	commatrix               *types.ComMatrix
-	isSNO                   bool
-	isBM                    bool
-	deployment              types.Deployment
-	infra                   types.Env
-	utilsHelpers            utils.UtilsInterface
-	epExporter              *endpointslices.EndpointSlicesExporter
-	nodeList                *corev1.NodeList
-	artifactsDir            string
+	cs           *client.ClientSet
+	commatrix    *types.ComMatrix
+	isSNO        bool
+	isBM         bool
+	deployment   types.Deployment
+	infra        types.Env
+	utilsHelpers utils.UtilsInterface
+	epExporter   *endpointslices.EndpointSlicesExporter
+	nodeList     *corev1.NodeList
+	artifactsDir string
 )
 
 const testNS = "openshift-commatrix-test"
--- a/test/e2e/e2e_suite_test.go
+++ b/test/e2e/e2e_suite_test.go
@@ -10,7 +10,6 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	"github.com/openshift-kni/commatrix/pkg/client"
-	commatrixcreator "github.com/openshift-kni/commatrix/pkg/commatrix-creator"
 	"github.com/openshift-kni/commatrix/pkg/endpointslices"
 	"github.com/openshift-kni/commatrix/pkg/types"
 	"github.com/openshift-kni/commatrix/pkg/utils"
@@ -28,8 +27,6 @@ var (
 	epExporter              *endpointslices.EndpointSlicesExporter
 	nodeList                *corev1.NodeList
 	artifactsDir            string
-	extraNFTablesMasterFile = ""
-	extraNFTablesWorkerFile = ""
 )
 
 const testNS = "openshift-commatrix-test"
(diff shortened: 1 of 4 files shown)


test/e2e/validation_test.go
```diff
@@ -153,7 +165,24 @@ var _ = Describe("Validation", func() {
 	})
 
 	It("should validate the communication matrix ports match the node's listening ports", func() {
-		err := commatrix.WriteMatrixToFileByType(utilsHelpers, "communication-matrix", types.FormatCSV, deployment, artifactsDir)
+		val, exists := os.LookupEnv("ENDPOINT_SLICE_EXTRA_OPEN_PORTS_FILE")
```
aabughosh, line 168: why not to do this on the before each?
will not need it on the nftables also? maybe we want to add ENDPOINT_SLICE_EXTRA_OPEN_PORTS_FORMAT

aabughosh:
maybe add example of that file on the e2e/examples

shirmoran:
> maybe add example of that file on the e2e/examples

Any file inside samples/custom-entries could be an example

test/e2e/validation_test.go
shirmoran: Those ports are open ports that does not have endpoint slices, so we only want to add them to the matrix in the test comparing endpoint slices and ss matrices - to avoid unnecessary failure.
In the other tests it will be meaning less.

test/e2e/nftables_test.go
```diff
@@ -27,35 +27,49 @@ var (
...
-					}
+				nftablesRules := roleMat.ToNFTablesRules()
+				Expect(err).NotTo(HaveOccurred())
+
+				extraNFTablesFile, _ := os.LookupEnv(extraNftablesFileEnv)
+				if extraNFTablesFile != "" {
+					extraNFTablesValue, err := os.ReadFile(extraNFTablesFile)
+					Expect(err).NotTo(HaveOccurred())
+					nftablesRules += "\n\t\t\t\t\t" + string(extraNFTablesValue)
+					Expect(err).NotTo(HaveOccurred())
```
aabughosh, line 65: this u dont need

test/e2e/nftables_test.go
```diff
@@ -27,35 +27,49 @@ var (
 var _ = Describe("Nftables", func() {
 	It("should apply firewall by blocking all ports except the ones OCP is listening on", func() {
 		masterMat, workerMat := commatrix.SeparateMatrixByRole()
+
+		var podsToIgnoreMasterMat, podsToIgnoreWorkerMat types.ComMatrix
+		if portsToIgnoreCommatrix != nil {
+			podsToIgnoreMasterMat, podsToIgnoreWorkerMat = portsToIgnoreCommatrix.SeparateMatrixByRole()
```
aabughosh, line 33: why the name is podsToIgnoreMasterMat amd not  portsToIgnoreMasterMat

test/e2e/nftables_test.go
```diff
@@ -27,35 +27,49 @@ var (
...
+					podsToIgnoreMat = podsToIgnoreMasterMat
+					extraNftablesFileEnv = "EXTRA_NFTABLES_MASTER_FILE"
+				}
 
-					if extraNFTablesMasterFile != "" {
-						nftablesConfig, err = AddPortsToNFTables(nftablesConfig, extraNFTablesMasterFile)
-						Expect(err).NotTo(HaveOccurred())
-					}
+				nftablesRules := roleMat.ToNFTablesRules()
+				Expect(err).NotTo(HaveOccurred())
```
aabughosh, line 58: this u dont need

test/e2e/nftables_test.go
shirmoran: right! thanks

test/e2e/e2e_suite_test.go
```diff
@@ -90,15 +94,21 @@ var _ = BeforeSuite(func() {
...
 
-	// get the EXTRA_NFTABLES_FILE if it not exist value is ""
-	val, exists := os.LookupEnv("EXTRA_NFTABLES_MASTER_FILE")
-	if exists {
-		extraNFTablesMasterFile = val
-	}
-
-	val, exists = os.LookupEnv("EXTRA_NFTABLES_WORKER_FILE")
+	portsToIgnoreCommatrix = nil
+	val, exists := os.LookupEnv("OPEN_PORTS_TO_IGNORE_FILE")
```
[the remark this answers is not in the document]
shirmoran: Those are open ports we want to ignore in the comparison to to the ss. We don't add them to the commatrix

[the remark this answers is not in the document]
aabughosh: In the nftable we are adding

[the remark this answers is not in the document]
shirmoran: True, though we're adding them because we want to ignore them to avoid failure.
Maybe this name is no accurate but I think "extra ports" could be miss leading because we don't document them as EPS.

PR #85, opened 2025-01-07
Allow specifying open ports to ignore in e2e documented commatrix test

shirmoran:
This PR allow specifying an extra `comdetails` file, that contains open ports that should be ignored when comparing the differences between the documented communication matrix with the endpoint slices communication matrix, to avoid unnecessary failures.
This option is needed for certain ports only.

The extra ports should be stored in a file (csv\json\yaml), and its path and format should be passed as the following env vars when running the e2e-test make target:
`OPEN_PORTS_TO_IGNORE_IN_DOC_TEST_FILE`, `OPEN_PORTS_TO_IGNORE_IN_DOC_TEST_FORMAT`

Additionally, this PR moves the extra nftables env vars extraction to the test's file.


test/e2e/nftables_test.go
```diff
@@ -34,25 +34,26 @@ var _ = Describe("Nftables", func() {
 			role, err := types.GetNodeRole(&node)
 			Expect(err).NotTo(HaveOccurred())
 
+			var roleMat types.ComMatrix
+			var extraNftablesFileEnv string
 			if _, exists := nodeRoleToNFTables[role]; !exists {
 				var nftablesConfig []byte
 
 				if role == workerNodeRole {
```
sabinaaledort, line 42: Can simplify:
```
extraNftablesFileEnv := EXTRA_NFTABLES_MASTER_FILE
if role == workerNodeRole {
    extraNftablesFileEnv = EXTRA_NFTABLES_WORKER_FILE
}
```

test/e2e/nftables_test.go
```diff
@@ -34,25 +34,26 @@ var _ = Describe("Nftables", func() {
...
+				}
+
+				nftablesConfig, err = roleMat.ToNFTables()
+				Expect(err).NotTo(HaveOccurred())
 
-					if extraNFTablesMasterFile != "" {
-						nftablesConfig, err = AddPortsToNFTables(nftablesConfig, extraNFTablesMasterFile)
-						Expect(err).NotTo(HaveOccurred())
-					}
+				extraNFTablesFile, _ := os.LookupEnv(extraNftablesFileEnv)
```
sabinaaledort, line 53: if extraNFTablesFile, ok := os.LookupEnv(extraNftablesFileEnv); ok {
 ......
}

test/e2e/validation_test.go
```diff
@@ -144,18 +140,35 @@ var _ = Describe("Validation", func() {
...
+			Expect(err).ToNot(HaveOccurred())
+			portsToIgnoreComDetails, err := portsToIgnoreCommatrixCreator.GetComDetailsListFromFile()
+			Expect(err).ToNot(HaveOccurred())
+			portsToIgnoreCommatrix = &types.ComMatrix{Matrix: portsToIgnoreComDetails}
+
+			// generate the diff matrix between the open ports to ignore matrix and the missing ports in the documented commatrix (based on the diff between the enpointslice and the doc matrix)
+			diffIgonre := matrixdiff.Generate(diff.GenerateUniquePrimary(), portsToIgnoreCommatrix)
+			missingPortsMat = diffIgonre.GenerateUniquePrimary()
+		} else {
+			missingPortsMat = diff.GenerateUniquePrimary()
```
sabinaaledort, line 162: Can be the default 
missingPortsMat := diff.GenerateUniquePrimary()

test/e2e/nftables_test.go
shirmoran: So the problem with that method, is that we pass an env var when we run the make target, so what happens is that if the env var is empty it just passes an empty string - and than the "ok" var will be assigned with true.
We can also change the way we send the env var when running the make target - check if the env var isn't empty, and just if so, send it.

sabinaaledort: You are right, let's keep the check

PR #101, opened 2025-03-06
Create cli executables for multiple platforms

bnshr:
This PR includes the following changes in the Makefile.

- Created commatrix cli executables in multiple platforms
- Added GOOS and GORACH to consider the current platform settings to create the executables

TODO: 
- Discuss how these executables would be used by the users

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,9 +20,47 @@ GOLANGCI_LINT = $(BIN_DIR)/golangci-lint
 # in case of a version bump
 GOLANGCI_LINT_VER = v1.63.4
 
+# Output directory
+BIN_DIR := bin
+
+# Supported platforms
+PLATFORMS := linux/amd64 linux/arm64 darwin/amd64 darwin/arm64 windows/amd64
+
+# Setting of GOOS and GOARCH for the build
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -20,9 +20,47 @@ GOLANGCI_LINT = $(BIN_DIR)/golangci-lint
 # in case of a version bump
 GOLANGCI_LINT_VER = v1.63.4
 
+# Output directory
+BIN_DIR := bin
```
aabughosh, line 24: u dont need this
we have it here:https://github.com/openshift-kni/commatrix/blob/45ce5346dde246bc49474cbcba18cdd9f1901fdf/Makefile#L14
this is why ur lint is failing

Makefile
```diff
@@ -20,9 +20,47 @@ GOLANGCI_LINT = $(BIN_DIR)/golangci-lint
 # in case of a version bump
 GOLANGCI_LINT_VER = v1.63.4
 
+# Output directory
+OUTPUT_DIR=$(CURPATH)/cross-build-output
```
aabughosh, line 24: can u add this file to the .gitignore file?

PR #109, opened 2025-03-17
Add the namespace and podname to the ss matrix output

aabughosh:
Add missing data to the ss matrix
namespace and pod


shirmoran (approved):

pkg/listening-sockets/listening-sockets.go
```diff
@@ -172,24 +174,24 @@ func (cc *ConnectionCheck) toComDetails(debugPod *corev1.Pod, ssOutput []string,
 	return res
 }
 
-// getContainerName receives an ss entry and gets the name of the container exposing this port.
-func (cc *ConnectionCheck) getContainerName(debugPod *corev1.Pod, ssEntry string) (string, error) {
+// getContaineInfo receives an ss entry and gets the info (containerName, podName, nameSpace) of the container exposing this port.
+func (cc *ConnectionCheck) getContaineInfo(debugPod *corev1.Pod, ssEntry string) (string, string, string, error) {
```
sabinaaledort, line 178: Missing 'r' in container

pkg/listening-sockets/listening_sockets_test.go
```diff
@@ -76,8 +76,8 @@ var (
 			Port:      111,
 			NodeRole:  "master",
 			Service:   "rpcbind",
-			Namespace: "",
-			Pod:       "",
+			Namespace: "test-namespace",
```
sabinaaledort, line 79: What's the reason for this change?

pkg/listening-sockets/listening-sockets.go
```diff
@@ -211,26 +213,28 @@ func (cc *ConnectionCheck) extractContainerID(debugPod *corev1.Pod, pid string)
 	return containerID, nil
 }
 
-// extractContainerName receives CRI-O container ID and returns the container's name.
-func (cc *ConnectionCheck) extractContainerName(debugPod *corev1.Pod, containerID string) (string, error) {
+// extractContainerName receives CRI-O container ID and returns the container's Info(containerName, containerNameSpace, containerPod).
+func (cc *ConnectionCheck) extractContainerInfo(debugPod *corev1.Pod, containerID string) (string, string, string, error) {
```
sabinaaledort, line 217: Maybe this func can return a container obj instead of returning a list of vars that might get bigger in the future if we decide to add additional container info

pkg/listening-sockets/listening_sockets_test.go
aabughosh: now we are returning the namespace and pod name and this will effect the test 
on the test we will have it

pkg/listening-sockets/listening-sockets.go
```diff
@@ -172,24 +174,24 @@ func (cc *ConnectionCheck) toComDetails(debugPod *corev1.Pod, ssOutput []string,
 	return res
 }
 
-// getContainerName receives an ss entry and gets the name of the container exposing this port.
-func (cc *ConnectionCheck) getContainerName(debugPod *corev1.Pod, ssEntry string) (string, error) {
+// getContaineInfo receives an ss entry and gets the info (containerName, podName, nameSpace) of the container exposing this port.
```
sabinaaledort, line 177: getContaineInfo -> getContainerInfo (missing r)

PR #110, opened 2025-03-18
Remove ports 5050 and 9447 from docs

shirmoran:
Since release 4.17 port 5050 is no longer open, and since release 4.18 port 9447 is no longer open.
Hence, they can be removed from the documented communication matrix in the corresponding versions.


sabinaaledort:
So these can be removed from the static entries as well right?

shirmoran:
> So these can be removed from the static entries as well right?

Right! Missed that part. I'll remove them.

sabinaaledort:
Maybe better separate each port to a PR so we can cherry-pick?

shirmoran:
> Maybe better separate each port to a PR so we can cherry-pick?

both of them has to get to 4.18 in this repo, 4.17 does not have the docs here, and will be removed from openshift-docs, and than we can remove it only from static entries in our repo.

PR #114, opened 2025-04-10
add a new required  flag for the platfrom type

aabughosh:
add a new required flag 
- added a new string var platform-type  (baremetal,aws). required.
- update the readme with the new flag and examples
- update the makefile to use the new var
- update the validation function to handle errors on the new var
- make it as required failed if not set
- update the unittest of the generate class
- change the infra var from Cloud to AWS
- add a missing compare to errors on the unitest

### DIFF
--- a/cmd/README.md
+++ b/cmd/README.md
@@ -43,6 +43,7 @@ Usage:
   oc commatrix generate [flags]
 
 Flags:
+      --platform-type string         Platform Type (baremetal, aws), Required
       --customEntriesFormat string   Set the format of the custom entries file (json,yaml,csv)
       --customEntriesPath string     Add custom entries from a file to the matrix
       --debug                        Debug logs (default is false)
@@ -58,10 +59,11 @@ Once you run the `oc commatrix generate` command, the plugin will
 generate a communication matrix based on the ingress flows in your
 OpenShift cluster. The output will be saved to a file (destDir) in the chosen format,
 similar to the following:
+> **Note:** The `--platform-type` flag is required for all commands.
 
-`csv example`
+`csv example on aws`
 ```sh
-$ oc commatrix generate --format csv
+$ oc commatrix generate --format csv --platform-type aws
 Direction,Protocol,Port,Namespace,Service,Pod,Container,Node Role,Optional
 Ingress,TCP,22,Host system service,sshd,,,master,true
 Ingress,TCP,53,openshift-dns,dns-default,dnf-default,dns,master,false
@@ -69,9 +71,9 @@ Ingress,TCP,80,openshift-ingress,router-internal-default,router-default,router,m
 Ingress,TCP,111,Host system service,rpcbind,,,master,true
 ```
 
-`json example`
+`json example on baremetal`
 ```sh
-$ oc commatrix generate --format json
+$ oc commatrix generate --format json --platform-type baremetal
 [
     {
         "direction": "Ingress",
@@ -98,9 +100,9 @@ $ oc commatrix generate --format json
 ]
 ```
 
-`host-open-ports example command`
+`host-open-ports example command on aws`
 ```sh
-$ oc commatrix generate --host-open-ports --format csv
+$ oc commatrix generate --platform-type aws --host-open-ports --format csv
 ```
 
 the command will generate the follwing paths:
@@ -151,9 +153,9 @@ UNCONN 0      0      127.0.0.1:323   0.0.0.0:* users:(("chronyd",pid=2805,fd=5))
 UNCONN 0      0      127.0.0.1:708   0.0.0.0:* users:(("rpc.statd",pid=3922,fd=8))
 ```
 
-`customEntriesFormat and customEntriesPath example command`
+`customEntriesFormat and customEntriesPath example command on baremetal`
 ```sh
-$ oc commatrix generate --format csv --customEntriesFormat csv --customEntriesPath "communication-matrix/customEntriesPath"
+$ oc commatrix generate --platform-type baremetal --format csv --customEntriesFormat csv --customEntriesPath "communication-matrix/customEntriesPath"
 ```
 
 `contents of communication-matrix/customEntriesPath`
--- a/cmd/generate/generate.go
+++ b/cmd/generate/generate.go
@@ -159,21 +159,33 @@ func Validate(o *GenerateOptions) error {
 			o.format, strings.Join(validFormats, ", "))
 	}
 
-	if o.customEntriesPath == "" {
-		return nil
-	}
-
-	if o.customEntriesFormat == "" {
-		return fmt.Errorf("you must specify the --customEntriesFormat when using --customEntriesPath")
-	}
-
-	if !slices.Contains(validCustomEntriesFormats, o.customEntriesFormat) {
-		return fmt.Errorf("invalid custom entries format '%s', valid options are: %s",
-			o.customEntriesFormat, strings.Join(validCustomEntriesFormats, ", "))
+	if err := validateCustomEntries(o.customEntriesPath, o.customEntriesFormat, validCustomEntriesFormats); err != nil {
+		return err
 	}
 
 	if !slices.Contains(validPlatformType, Platform(o.platformType)) {
 		return fmt.Errorf("invalid platform type %s, valid options: %v", o.platformType, validPlatformType)
+	}
+
+	return nil
+}
+
+func validateCustomEntries(path, format string, validFormats []string) error {
+	if path == "" && format == "" { // dont need to validate
+		return nil
+	}
+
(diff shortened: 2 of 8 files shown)


pkg/utils/utils.go
```diff
@@ -236,6 +236,11 @@ func (u *utils) IsBMInfra() (bool, error) {
 		return false, err
 	}
 
+	if infra.Status.PlatformStatus.Type == configv1.NonePlatformType {
+		log.Println("Detected platform type as 'None'; assuming cloud by default. " +
```
sabinaaledort, line 240: This will be printed only if the logger is enabled?
Should this be printed as a warning?
Should utils print user messages? I think this print should be done in cmd

cmd/generate/generate.go
```diff
@@ -109,6 +110,13 @@ func NewCmdCommatrixGenerate(cs *client.ClientSet, streams genericiooptions.IOSt
 		Long:    commatrixLong,
 		Example: CommatrixExample,
 		RunE: func(c *cobra.Command, args []string) (err error) {
+			if !c.Flags().Changed("isBM") {
```
sabinaaledort, line 113: Can this be done by using the flag var? To avoid the hardcoded string

cmd/README.md
```diff
@@ -49,6 +49,7 @@ Flags:
       --destDir string               Output files dir (default communication-matrix)
       --format string                Desired format (json,yaml,csv,nft) (default "csv")
       --host-open-ports              Generate communication matrix, host open port matrix, and their difference.
+      --isBM                         Platform is baremetal (optional;auto detect if not set).
```
oribon, line 52: not sure I understand the reason for the flag, if we can "auto detect" why would the user choose to mention it explicitly?
can you write the reason in the commit message as well + its title should be shorter e.g. "add .." without the "if it not set code will detect it"

cmd/generate/generate.go
```diff
@@ -123,12 +131,17 @@ func NewCmdCommatrixGenerate(cs *client.ClientSet, streams genericiooptions.IOSt
...
 	cmd.Flags().StringVar(&o.format, "format", "csv", "Desired format (json,yaml,csv,nft)")
-	cmd.Flags().BoolVar(&o.debug, "debug", false, "Debug logs")
+	cmd.Flags().StringVar(&o.platformType, "platform-type", "", "Platform Type (baremetal, aws). Required")
 	cmd.Flags().StringVar(&o.customEntriesPath, "customEntriesPath", "", "Add custom entries from a file to the matrix")
 	cmd.Flags().StringVar(&o.customEntriesFormat, "customEntriesFormat", "", "Set the format of the custom entries file (json,yaml,csv)")
 	cmd.Flags().BoolVar(&o.openPorts, "host-open-ports", false, "Generate communication matrix, host open ports matrix, and their difference")
+	cmd.Flags().BoolVar(&o.debug, "debug", false, "Debug logs")
+	if err := cmd.MarkFlagRequired("platform-type"); err != nil { // make it as required flag
+		log.Fatalf("failed to mark platform-type as required: %v", err)
+	}
```
oribon, line 144: I think this can be removed? we have the check beneath

cmd/generate/generate.go
```diff
@@ -139,15 +152,16 @@ func Validate(o *GenerateOptions) error {
...
+			strings.Join(validPlatformType, ", "))
 	}
 
-	if o.customEntriesFormat == "" {
+	if o.customEntriesPath != "" && o.customEntriesFormat == "" {
 		return fmt.Errorf("you must specify the --customEntriesFormat when using --customEntriesPath")
 	}
 
-	if !slices.Contains(validCustomEntriesFormats, o.customEntriesFormat) {
+	if o.customEntriesFormat != "" && !slices.Contains(validCustomEntriesFormats, o.customEntriesFormat) {
```
oribon, line 164: why this change?

cmd/generate/generate.go
```diff
@@ -139,15 +152,16 @@ func Validate(o *GenerateOptions) error {
 			o.format, strings.Join(validFormats, ", "))
 	}
 
-	if o.customEntriesPath == "" {
-		return nil
+	if !slices.Contains(validPlatformType, strings.ToLower(o.platformType)) {
+		return fmt.Errorf("invalid platform type %s, please use one of the following: %s", o.platformType,
+			strings.Join(validPlatformType, ", "))
```
oribon, line 157: can you put this the lowest?

cmd/generate/generate.go
```diff
@@ -189,12 +203,7 @@ func Run(o *GenerateOptions) (err error) {
 		deployment = types.SNO
 	}
 
-	isBM, err := o.utilsHelpers.IsBMInfra()
-	if err != nil {
-		return fmt.Errorf("failed to check is bm cluster %s", err)
-	}
-
-	if isBM {
+	if strings.ToLower(o.platformType) == validPlatformType[0] { // validPlatformType[0]="baremetal"
```
oribon, line 206: no need for the toLower, and match it with the const (from the previous comment) instead of using slice[0]

cmd/generate/generate.go
oribon, line 1: I see we still have types.Cloud, wdyt about renaming it to AWS?

cmd/generate/generate.go
aabughosh: this if was related so i made them to unrelated to each other
if u realize they was one after one

cmd/generate/generate.go
aabughosh: i want to separate the error msg if it invalid or missing

cmd/generate/generate.go
```diff
@@ -139,15 +152,16 @@ func Validate(o *GenerateOptions) error {
...
 
-	if o.customEntriesPath == "" {
-		return nil
+	if !slices.Contains(validPlatformType, strings.ToLower(o.platformType)) {
+		return fmt.Errorf("invalid platform type %s, please use one of the following: %s", o.platformType,
+			strings.Join(validPlatformType, ", "))
 	}
 
-	if o.customEntriesFormat == "" {
+	if o.customEntriesPath != "" && o.customEntriesFormat == "" {
```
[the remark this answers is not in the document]
aabughosh: before if u see the if's was related so i made them to be not releated to what we had before

cmd/generate/generate.go
oribon: this is unrelated to the "platform type" change, this should go in a separate commit

cmd/generate/generate.go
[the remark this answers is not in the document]
oribon: same (should be in a separate commit)

cmd/generate/generate.go
```diff
@@ -159,15 +159,15 @@ func Validate(o *GenerateOptions) error {
 			o.format, strings.Join(validFormats, ", "))
 	}
 
-	if o.customEntriesPath == "" {
-		return nil
+	if o.customEntriesPath != "" && o.customEntriesFormat == "" {
```
oribon, line 162: let's wrap everything related to customEntries in a validateCustomEntries func, and return nil first thing when they are both empty.

cmd/generate/generate.go
```diff
@@ -159,15 +159,15 @@ func Validate(o *GenerateOptions) error {
...
 	}
 
-	if o.customEntriesFormat == "" {
-		return fmt.Errorf("you must specify the --customEntriesFormat when using --customEntriesPath")
+	if o.customEntriesPath == "" && o.customEntriesFormat != "" {
+		return fmt.Errorf("you must specify the --customEntriesPath when using --customEntriesFormat")
 	}
 
-	if !slices.Contains(validCustomEntriesFormats, o.customEntriesFormat) {
+	if o.customEntriesFormat != "" && !slices.Contains(validCustomEntriesFormats, o.customEntriesFormat) {
```
oribon, line 170: when you return nil on both empty, you won't need the `o.customEntriesFormat != ""`

cmd/generate/generate_test.go
```diff
@@ -165,34 +165,50 @@ func TestCommatrixGeneration(t *testing.T) {
...
 			name: "Should Return failure on format validation",
-			args: []string{"generate", "--format", "test"},
+			args: []string{"generate", "--format", "test", "--platform-type", "aws"},
 			expectedFunc: func() (string, error) {
 				return "", fmt.Errorf("invalid format 'test', valid options are: csv, json, yaml, nft")
 			},
 			wantErr: true,
 		},
 		{
 			name: "Should return failure when customEntriesPath is set but customEntriesFormat is missing",
```
oribon, line 200: should we add a case for customEntriesPath missing but customEntriesFormat set?

oribon (approved):
/lgtm

PR #115, opened 2025-04-22
refactor e2e test

aabughosh:
- use the oc command to generate the matric's and read them, for simulate for user running.
- Create a `ParseToComDetailsList` function to refactor the process of loading `ComDetailsList` from files, removing reliance on the `commMatrixCreator`.
- This refactor improves the e2e test by making it closer to user behavior and centralizing the parsing logic for better reuse and maintainability.
- Replaced internal matrix generation functions with the OpenShift CLI (oc) commands to generate and retrieve matrices, aligning the test process with actual user interactions.

### DIFF
--- a/test/e2e/nftables_test.go
+++ b/test/e2e/nftables_test.go
@@ -1,9 +1,11 @@
 package e2e
 
 import (
+	"context"
 	"fmt"
 	"log"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strings"
 
@@ -19,17 +21,39 @@ import (
 )
 
 var (
-	workerNodeRole = "worker"
-	tableName      = "table inet openshift_filter"
-	chainName      = "chain OPENSHIFT"
+	workerNodeRole     = "worker"
+	tableName          = "table inet openshift_filter"
+	chainName          = "chain OPENSHIFT"
+	workerNFTFile      = "communication-matrix-worker.nft"
+	masterNFTFile      = "communication-matrix-master.nft"
+	masterNFTconfig    []byte
+	workerNFTconfig    []byte
+	nodeRoleToNFTables map[string][]byte
 )
 
 var _ = Describe("Nftables", func() {
-	It("should apply firewall by blocking all ports except the ones OCP is listening on", func() {
-		masterMat, workerMat := commatrix.SeparateMatrixByRole()
-		nodeRoleToNFTables := make(map[string][]byte)
+	BeforeEach(func() {
+		By("Creating test namespace")
+		err := utilsHelpers.CreateNamespace(testNS)
+		Expect(err).ToNot(HaveOccurred())
 
+		nodeList = &corev1.NodeList{}
+		err = cs.List(context.TODO(), nodeList)
+		Expect(err).ToNot(HaveOccurred())
+
+		By("Generating nft communication matrix using oc command")
+		cmd := exec.Command("oc", "commatrix", "generate", "--format", "nft", "--destDir", artifactsDir)
+		err = cmd.Run()
+		Expect(err).ToNot(HaveOccurred(), fmt.Sprintf("Failed to run command: %s", cmd.String()))
+
+		By("Reading the generated commatrix nft files")
+		masterFilePath := filepath.Join(artifactsDir, masterNFTFile)
+		masterNFTconfig, err = os.ReadFile(masterFilePath)
+		Expect(err).ToNot(HaveOccurred(), fmt.Sprintf("Failed to read generated %s file", masterNFTFile))
+
+		// creare the nodeRoleToNFTables map[string][]byte
 		By("Creating NFT output for each role")
+		nodeRoleToNFTables = make(map[string][]byte)
 		for _, node := range nodeList.Items {
 			role, err := types.GetNodeRole(&node)
 			Expect(err).NotTo(HaveOccurred())
@@ -37,15 +61,16 @@ var _ = Describe("Nftables", func() {
 			if _, exists := nodeRoleToNFTables[role]; !exists {
 				var nftablesConfig []byte
 
-				roleBasedMat := masterMat
+				nftablesConfig = masterNFTconfig
 				extraNftablesFileEnv := "EXTRA_NFTABLES_MASTER_FILE"
 				if role == workerNodeRole {
-					roleBasedMat = workerMat
+					workerFilePath := filepath.Join(artifactsDir, workerNFTFile)
+					workerNFTconfig, err = os.ReadFile(workerFilePath)
+					Expect(err).ToNot(HaveOccurred(), fmt.Sprintf("Failed to read generated %s file", workerNFTFile))
+
+					nftablesConfig = workerNFTconfig
 					extraNftablesFileEnv = "EXTRA_NFTABLES_WORKER_FILE"
 				}
-
-				nftablesConfig, err = roleBasedMat.ToNFTables()
-				Expect(err).NotTo(HaveOccurred())
 
 				extraNFTablesFile, _ := os.LookupEnv(extraNftablesFileEnv)
 				if extraNFTablesFile != "" {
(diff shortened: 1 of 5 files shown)


bnshr (approved):

test/e2e/validation_test.go
```diff
@@ -254,3 +261,41 @@ func filterSSMatrix(mat *types.ComMatrix) (*types.ComMatrix, error) {
...
+
+	missingEPSMat, err := types.ParseToComDetailsList([]byte(minusLines), types.FormatCSV)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return &types.ComMatrix{Matrix: notUsedEPSMat}, &types.ComMatrix{Matrix: missingEPSMat}, nil
+}
+
+func filterCSVPlusMinusLines(csvContent string, prefix string) string {
```
bnshr, line 282: A better name could be `filterCsvLinesByPrefix` as this code is generic.

aabughosh: Seems better

shirmoran, line 282: Suggestion for an alternative func name: `extractDiffByPrefix`

test/e2e/validation_test.go
```diff
@@ -254,3 +261,41 @@ func filterSSMatrix(mat *types.ComMatrix) (*types.ComMatrix, error) {
 
 	return &types.ComMatrix{Matrix: res}, nil
 }
+
+func extractNotUsedAndMissingEPS(csvContent string) (*types.ComMatrix, *types.ComMatrix, error) {
+	plusLinesCSV := filterCSVPlusMinusLines(csvContent, "+")
+	minusLines := filterCSVPlusMinusLines(csvContent, "-")
+
+	notUsedEPSMat, err := types.ParseToComDetailsList([]byte(plusLinesCSV), types.FormatCSV)
```
bnshr, line 269: unused instead of notUsed.

test/e2e/validation_test.go
```diff
@@ -49,17 +47,40 @@ var (
...
+			cmd := exec.Command("oc", "commatrix", "generate", "--host-open-ports", "--destDir", artifactsDir)
+			err := cmd.Run()
+			Expect(err).ToNot(HaveOccurred(), "Failed to run oc commatrix generate")
+
+			By("Reading the generated commatrix files")
+			commatrixFilePath := filepath.Join(artifactsDir, commatrixFile)
+			commatrixFileContent, err := os.ReadFile(commatrixFilePath)
+			Expect(err).ToNot(HaveOccurred(), "Failed to read generated commatrix file")
+
+			ComDetailsCommatrix, err := types.ParseToComDetailsList(commatrixFileContent, types.FormatCSV)
```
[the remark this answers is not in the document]
aabughosh: comMatrix is better since ComMatrix is a list of ComDetails and this value is ComDetails list
type ComMatrix struct {
	Matrix []ComDetails
}

test/e2e/validation_test.go
aabughosh: i use the same name as before what it was

test/e2e/validation_test.go
```diff
@@ -97,17 +118,18 @@ var _ = Describe("Validation", func() {
...
 		By("generating diff between matrices for testing purposes")
 		endpointslicesDiffWithDocMat := matrixdiff.Generate(commatrix, docComMatrix)
 		diffStr, err := endpointslicesDiffWithDocMat.String()
 		Expect(err).ToNot(HaveOccurred())
+
 		err = os.WriteFile(filepath.Join(artifactsDir, "doc-diff-new"), []byte(diffFileComments+diffStr), 0644)
 		Expect(err).ToNot(HaveOccurred())
 
 		By("comparing new and documented commatrices")
-		// Get ports that are in the documented commatrix but not in the generated commatrix.
```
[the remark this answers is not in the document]
aabughosh: sorry by mistake i will return it back

test/e2e/validation_test.go
```diff
@@ -254,3 +261,41 @@ func filterSSMatrix(mat *types.ComMatrix) (*types.ComMatrix, error) {
 
 	return &types.ComMatrix{Matrix: res}, nil
 }
+
+func extractNotUsedAndMissingEPS(csvContent string) (*types.ComMatrix, *types.ComMatrix, error) {
```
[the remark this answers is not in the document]
aabughosh: its extracting the not used endpointslice 
like endpoint slice is there but the ports of this endpint slice are not in use so not sure about this name

test/e2e/validation_test.go
```diff
@@ -49,17 +47,40 @@ var (
...
 	diffFileComments         = "// `+` indicates a port that isn't in the current documented matrix, and has to be added.\n" +
 		"// `-` indicates a port that has to be removed from the documented matrix.\n"
+	commatrixFile      = "communication-matrix.csv"
+	matrixdiffFile     = "matrix-diff-ss"
 	serviceNodePortMin = 30000
 	serviceNodePortMax = 32767
 )
 
 var _ = Describe("Validation", func() {
+	runeOnce := false
```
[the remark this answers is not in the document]
shirmoran: sounds good

test/e2e/validation_test.go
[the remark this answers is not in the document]
shirmoran: What do you mean by the endpointslice is "there"?
The endpoint slice is documented, but the **port is not used** (meaning an endpointslice for this port also does not exist).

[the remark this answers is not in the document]
aabughosh: like we get the endpint slice (not related if it documented or not) but on the system the port that this edpoint slice have is unused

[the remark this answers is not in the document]
aabughosh: This is not the documented test by the way
Its the one that comparing between ss output and endpointslice list output 
I think u are might mixing

test/e2e/nftables_test.go
```diff
@@ -122,6 +148,12 @@ var _ = Describe("Nftables", func() {
 			Fail("OPENSHIFT chain not found in nftables")
 		}
 	})
+
+	AfterEach(func() {
```
oribon, line 152: nit: put this below the BeforeEach

test/e2e/nftables_test.go
```diff
@@ -12,41 +14,65 @@ import (
...
 				var nftablesConfig []byte
 
-				roleBasedMat := masterMat
+				nftablesConfig = masterNFTconfig
 				extraNftablesFileEnv := "EXTRA_NFTABLES_MASTER_FILE"
 				if role == workerNodeRole {
-					roleBasedMat = workerMat
+					workerFilePath := filepath.Join(artifactsDir, workerNFTFile)
+					workerNFTconfig, err = os.ReadFile(workerFilePath)
+					Expect(err).ToNot(HaveOccurred(), fmt.Sprintf("Failed to read generated %s file", workerNFTFile))
```
oribon, line 70: can we do this in the BeforeEach?

test/e2e/validation_test.go
```diff
@@ -49,17 +47,40 @@ var (
...
+	commatrixFile      = "communication-matrix.csv"
+	matrixdiffFile     = "matrix-diff-ss"
 	serviceNodePortMin = 30000
 	serviceNodePortMax = 32767
 )
 
 var _ = Describe("Validation", func() {
+	commatrixGenerated := false
+	BeforeEach(func() {
+		if !commatrixGenerated {
```
oribon, line 66: what's the reason not to create it everytime?
also, this can just be
```go
if commatrix != nil {
  return
}
...
```

test/e2e/validation_test.go
```diff
@@ -254,3 +262,41 @@ func filterSSMatrix(mat *types.ComMatrix) (*types.ComMatrix, error) {
...
+
+	missingEPSMat, err := types.ParseToComDetailsList([]byte(missingEPS), types.FormatCSV)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return &types.ComMatrix{Matrix: notUsedEPSMat}, &types.ComMatrix{Matrix: missingEPSMat}, nil
+}
+
+func extractDiffByPrefix(csvContent string, prefix string) string {
```
oribon, line 283: please add a description here (with an example). also, wdyt about having an enum instead of string? I feel like this is not really about prefixes, but about "types" (missing / unused / etc.)

test/e2e/validation_test.go
```diff
@@ -254,3 +262,41 @@ func filterSSMatrix(mat *types.ComMatrix) (*types.ComMatrix, error) {
 
 	return &types.ComMatrix{Matrix: res}, nil
 }
+
+func extractNotUsedAndMissingEPS(csvContent string) (*types.ComMatrix, *types.ComMatrix, error) {
```
oribon, line 266: can we split this into 2 clear functions? not sure I see the benefit of combining them, would be cleaner having one function for unused and one for missing returning a matrix, and let the caller call w.e they want

test/e2e/validation_test.go
aabughosh: Its 2 tests using same base 
No need to run the command twice

PR #120, opened 2025-05-04
Update commatrix version

aabughosh:
- update the commatrix version on the krew.yaml to 0.0.2
- fix the adding of the platfor-type var to the test command releated to this pr:https://github.com/openshift-kni/commatrix/pull/114


test/e2e/e2e_suite_test.go
```diff
@@ -48,8 +52,15 @@ var _ = BeforeSuite(func() {
...
 	Expect(err).NotTo(HaveOccurred())
 
-	isBM, err = utilsHelpers.IsBMInfra()
-	Expect(err).NotTo(HaveOccurred())
+	platformType, _ = os.LookupEnv("PLATFORM_TYPE")
+	if platformType == "" {
+		platformType = string(generate.PlatformAWS)
+	}
+
+	isBM = false
```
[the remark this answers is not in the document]
aabughosh: its for all not just to the doc

test/e2e/e2e_suite_test.go
```diff
@@ -48,8 +52,22 @@ var _ = BeforeSuite(func() {
 	isSNO, err = utilsHelpers.IsSNOCluster()
 	Expect(err).NotTo(HaveOccurred())
 
-	isBM, err = utilsHelpers.IsBMInfra()
-	Expect(err).NotTo(HaveOccurred())
+	isBM = false
+	platformType, _ = os.LookupEnv("PLATFORM_TYPE")
+	if platformType == string(generate.PlatformBaremetal) { // if it set
```
shirmoran, line 57: This check does not say the platform type is set, it says it's sell to **bm**.
You should check the error from the `os.LookupEnv("PLATFORM_TYPE")`, if there is an error it's not set.
Also, if `PLATFORM_TYPE` is set to an invalid type (not bm ar aws), the test should either fail or continue with a msg saying the var is not set and we are using a different value.
We need to decide what is out "default" reaction for when the var is not set.

test/e2e/e2e_suite_test.go
```diff
@@ -48,8 +52,23 @@ var _ = BeforeSuite(func() {
...
+		isBM = true
+	}
+
+	if platformType == "" || (platformType != string(generate.PlatformBaremetal) && platformType != string(generate.PlatformAWS)) { // if not set use IsBMInfra function
```
shirmoran, line 61: I would move this check above `if platformType == string(generate.PlatformBaremetal)`

shirmoran:
LGTM on my side.

PR #121, opened 2025-05-06
export GetStaticEntries for external use and add GetCommonEntries function on diff package

aabughosh:
- export GetStaticEntries for external use
  the getStaticEntries function has been exported (made public) by renaming it to GetStaticEntries, allowing it to be used in other package
- create a GetCommonEntries to return the common entries on a matrix
  Create GetCommonEntries to return entries shared by both matrices in a diff
- Renamed GenerateUniquePrimary and GenerateUniqueSecondary to GetUniquePrimary and GetUniqueSecondary for better clarity, since the functions return values rather than generate the


bnshr (approved):

oribon (commented):
/lgtm
/approve

PR #124, opened 2025-05-13
handle the noneplatformtype platform (removing the added platform-type var and add the support of none in the code)

aabughosh:
We originally introduced the PLATFORM_TYPE environment variable to handle cases where the platform type was None. However, None is a valid platform type on its own and should not be interpreted as either a cloud or baremetal platform.

Therefore, we are removing all usage of the PLATFORM_TYPE variable and will rely solely on the existing functions that detect the platform type automatically.

update the code to handle the none platfrom type:
- on case of test need to compare the none sno to the bm sno
- on the static entries no need to add for none
- remove the env iota var and use the configv1 var
- supported telco cases for now are (bm sno thats none and bm mno and the aws)

change the d var to be deployment a clear name

### DIFF
--- a/cmd/generate/generate.go
+++ b/cmd/generate/generate.go
@@ -33,31 +33,22 @@ var (
 
 	`)
 	CommatrixExample = templates.Examples(`
-			 # Note: The --platform-type flag is required for all commands.
-
-			 # Generate the communication matrix in default format (csv) on AWS:
-			 oc commatrix generate --platform-type aws 
+			 # Generate the communication matrix in default format (csv):
+			 oc commatrix generate  
 			 
-			 # Generate the communication matrix in nft format on baremetal platform:
-			 oc commatrix generate --platform-type baremetal --format nft 
-
-			 # Generate the communication matrix in json format with debug logs on AWS:
-			 oc commatrix generate --platform-type aws --format json --debug
+			 # Generate the communication matrix in nft format:
+			 oc commatrix generate --format nft 
 			 
-			 # Generate communication matrix, host open ports matrix, and their difference in yaml format on baremetal:
-			 oc commatrix generate --platform-type baremetal --host-open-ports --format yaml 
+			 # Generate the communication matrix in json format with debug logs:
+			 oc commatrix generate --format json --debug
 			 
-			 # Generate the communication matrix in json format with custom entries on AWS:
-			 oc commatrix generate --platform-type aws --format json --customEntriesPath /path/to/customEntriesFile --customEntriesFormat json
+			 # Generate communication matrix, host open ports matrix, and their difference in yaml format:
+			 oc commatrix generate --host-open-ports --format yaml 
+			 
+			 # Generate the communication matrix in json format with custom entries:
+			 oc commatrix generate --format json --customEntriesPath /path/to/customEntriesFile --customEntriesFormat json
 			 
 	`)
-)
-
-type Platform string
-
-const (
-	PlatformBaremetal Platform = "baremetal"
-	PlatformAWS       Platform = "aws"
 )
 
 var (
@@ -73,15 +64,9 @@ var (
 		types.FormatJSON,
 		types.FormatYAML,
 	}
-
-	validPlatformType = []Platform{
-		PlatformBaremetal,
-		PlatformAWS,
-	}
 )
 
 type GenerateOptions struct {
-	platformType        string
 	destDir             string
 	format              string
 	customEntriesPath   string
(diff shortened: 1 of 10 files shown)


cmd/generate/generate.go
```diff
@@ -225,7 +203,12 @@ func Run(o *GenerateOptions) (err error) {
 		deployment = types.SNO
 	}
 
-	if Platform(o.platformType) == PlatformBaremetal {
+	platformType, err := o.utilsHelpers.GetPlatformType()
+	if err != nil {
+		return fmt.Errorf("failed to check is sno cluster %s", err)
```
bnshr, line 208: Is it typo? "is" instead of "if"

cmd/generate/generate.go
```diff
@@ -225,10 +203,19 @@ func Run(o *GenerateOptions) (err error) {
 		deployment = types.SNO
 	}
 
-	if Platform(o.platformType) == PlatformBaremetal {
+	platformType, err := o.utilsHelpers.GetPlatformType()
+	if err != nil {
+		return fmt.Errorf("failed to get platform type %s", err)
+	}
+
+	if platformType == configv1.BareMetalPlatformType {
```
shirmoran, line 211: Maybe we should initialize the `infra` variable above this line? So readers won't have to look where this variable is intialized.
Same regarding `deployment` var.

pkg/types/types.go
```diff
@@ -25,6 +25,7 @@ type Env int
 const (
 	Baremetal Env = iota
 	AWS
+	NonePlatformType
```
shirmoran, line 28: Do we even need those consts? maybe we can just save the infra type using the `configv1`  consts as we return them from the `GetPlatformType()` func.

pkg/utils/utils.go
```diff
@@ -229,14 +229,14 @@ func (u *utils) IsSNOCluster() (bool, error) {
 	return infra.Status.ControlPlaneTopology == configv1.SingleReplicaTopologyMode, nil
 }
 
-func (u *utils) IsBMInfra() (bool, error) {
+func (u *utils) GetPlatformType() (configv1.PlatformType, error) {
```
shirmoran, line 232: Maybe we should return an error if the platform type is not supported (not none, bm or aws)?

test/e2e/e2e_suite_test.go
```diff
@@ -48,8 +49,13 @@ var _ = BeforeSuite(func() {
 	isSNO, err = utilsHelpers.IsSNOCluster()
 	Expect(err).NotTo(HaveOccurred())
 
-	isBM, err = utilsHelpers.IsBMInfra()
+	platformType, err := utilsHelpers.GetPlatformType()
 	Expect(err).NotTo(HaveOccurred())
+
+	isBM = false
+	if platformType == configv1.BareMetalPlatformType || platformType == configv1.NonePlatformType { // if it mno bm or sno bm (sno bm is with noneplatfrom type)
```
shirmoran, line 56: I think we should add another comments saying that we **assume** telco partners use none platform type in bm only.

pkg/types/types.go
aabughosh: we dont need we can change all our ising of this and just used the platfromtype of configv1

pkg/utils/utils.go
shirmoran: There other platforms are valid but they are not supported, so if someone will try to run the test we another platform type it will fail as it will assume to be aws.
This is why returning an error saying the platform is **not supported** in our application is a good solution in my opinion.

aabughosh: i added it at the end since on our doc we are just talking about that supported cases if we will support more on the future so we need to update the slice

pkg/commatrix-creator/commatrix.go
```diff
@@ -10,22 +10,23 @@ import (
...
+	configv1 "github.com/openshift/api/config/v1"
 )
 
 type CommunicationMatrixCreator struct {
 	exporter            *endpointslices.EndpointSlicesExporter
 	customEntriesPath   string
 	customEntriesFormat string
-	e                   types.Env
+	platformType        configv1.PlatformType
 	d                   types.Deployment
```
shirmoran, line 21: as we changed `e` to `platformType`, maybe we should change `d` to `deployment` for consistency matters?

pkg/types/types.go
```diff
@@ -67,17 +60,6 @@ type ContainerInfo struct {
...
-	case "baremetal":
-		return Baremetal, nil
-	case "aws":
-		return AWS, nil
-	default:
-		return -1, fmt.Errorf("invalid cluster environment: %s", envStr)
-	}
-}
-
 func GetDeployment(deploymentStr string) (Deployment, error) {
```
shirmoran, line 63: I think this function is also not used.

oribon:
is it possible to refactor the pr so that it reverts the previous commit (`git-revert`), and add the new changes in a different commit?

PR #125, opened 2025-05-20
Remove ports 10300 and 10390 from doc commatrix and static entries

shirmoran:
Ports 10300, 10390 had been changed to be used only internally and there for shouldn't be documented in the communication matrix (see resolved [JIRA bug](https://issues.redhat.com/browse/OCPBUGS-50493), and [PR](https://github.com/openshift/csi-operator/pull/364) converting the ports to internal use)

Note: `types.CloudStaticEntriesWorker` was deleted as it only contained the above ports.

### DIFF
--- a/pkg/commatrix-creator/commatrix_test.go
+++ b/pkg/commatrix-creator/commatrix_test.go
@@ -240,8 +240,8 @@ var _ = g.Describe("Commatrix creator pkg tests", func() {
 			o.Expect(err).ToNot(o.HaveOccurred())
 
 			g.By("Comparing gotten static entries to wanted comDetails")
-			wantedComDetails := slices.Concat(types.CloudStaticEntriesMaster, types.CloudStaticEntriesWorker,
-				types.GeneralStaticEntriesMaster, types.StandardStaticEntries, types.GeneralStaticEntriesWorker)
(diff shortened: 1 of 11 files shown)


pkg/commatrix-creator/commatrix_test.go
```diff
@@ -239,8 +239,8 @@ var _ = g.Describe("Commatrix creator pkg tests", func() {
 			o.Expect(err).ToNot(o.HaveOccurred())
 
 			g.By("Comparing gotten static entries to wanted comDetails")
-			wantedComDetails := slices.Concat(types.CloudStaticEntriesMaster, types.CloudStaticEntriesWorker,
-				types.GeneralStaticEntriesMaster, types.StandardStaticEntries, types.GeneralStaticEntriesWorker)
+			wantedComDetails := slices.Concat(types.CloudStaticEntriesMaster, types.GeneralStaticEntriesMaster,
```
aabughosh, line 242: can u add more explanation about that?
or maybe to saperate into to commits one for removing the ports and the other is to remove this since there was just 2 that u removed on this var and this is why u deleted it

shirmoran: Will add a note in the description regarding that.

aabughosh:
looks good to me

oribon (commented):
add the comment from the description to the commit message, other than that lgtm

PR #128, opened 2025-05-27
[release-4.18] handle the noneplatformtype platform on 4.18

aabughosh:
- remove the itoa env var, no need for it and use the configv1 var 
- update the function of the platform type to get the type and if not suppourted to show relevant error 
- ensured that static entries are not created for None platform type. 
- update the deployment var name from d to be deployment, to consistenty with the other paramters and remove the unused function
cherry pick from https://github.com/openshift-kni/commatrix/pull/124

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -63,14 +63,9 @@ func main() {
 	}
 
 	log.Debug("Get infra type")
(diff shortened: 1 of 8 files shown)


bnshr (approved):
lgtm

PR #137, opened 2025-05-29
improve error message on e2e test and add a waiting for namespace

aabughosh:
capture stdout and stderr for failed command on the e2e test that will help also to debug issues on the command


SchSeba (approved):
/lgtm
/approve

PR #139, opened 2025-05-29
Change bm-sno commatrix name to none-sno

shirmoran:
Since the commatrix documented from baremetal sno clusters is of platform type none, we decided to change the name of the file to avoid confusion (as the platform type is not baremetal).


test/e2e/validation_test.go
```diff
@@ -61,11 +61,13 @@ var _ = Describe("Validation", func() {
 	It("generated communication matrix should be equal to documented communication matrix", func() {
 		By("generate documented commatrix file path")
 		docType := "aws"
```
yuvalk, line 63: can we change the logic here entirely so we wont assume aws as default?
https://github.com/openshift/api/blob/a463917701ab4506585893739e576ffd1e6b924e/config/v1/types_infrastructure.go#L410
we need to cover all the cases here, where we should error for platforms we dont support.

so this should be a simple switch case, which defults to not supported

yuvalk:
/approve

working incrementally here
and @shirmoran will open a separate PR to fix the switch..case

PR #140, opened 2025-05-29
add a new none documentation for mno cluster

aabughosh:
since that RAN can install baremetal on nonetypeplatfrom on mno so we are adding the documentation of them


aabughosh:
> what's between the commit msg and the actual content? I think you are adding outcome files for platform none? where is the mentioned test case?

by mistake o wrote test im channing

PR #144, opened 2025-06-03
Improve supported cluster's platform type check

shirmoran:
Instead of checking if the cluster's platform type is supported within the `GetPlatformType` function, we will check and handle the platform type out side of the function to avoid the assumption the function is always called in order to get the platform type.

Also if cluster's platform type is not supported, the e2e tests will be
skipped.

### DIFF
--- a/test/e2e/e2e_suite_test.go
+++ b/test/e2e/e2e_suite_test.go
@@ -1,14 +1,17 @@
 package e2e
 
 import (
+	"fmt"
 	"log"
 	"os"
 	"path/filepath"
+	"slices"
 	"testing"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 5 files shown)


test/e2e/e2e_suite_test.go
```diff
@@ -48,9 +46,6 @@ var _ = BeforeSuite(func() {
 	utilsHelpers = utils.New(cs)
 	isSNO, err = utilsHelpers.IsSNOCluster()
 	Expect(err).NotTo(HaveOccurred())
-
-	platformType, err = utilsHelpers.GetPlatformType()
```
aabughosh, line 52: maybe better to keep it here? and to run the error from here?
the e2e test is calling the oc generate command so if there is any error on the platform type the oc generate will catch from there and that will not reach ur error there

shirmoran: I also edited the `generate` file so it will be caught.

PR #145, opened 2025-06-04
[release-4.17] Remove port 9537 from static entries

shirmoran:
Port 9537 is not exposed and hence was removed from documented commatrix (see [PR#93895](https://github.com/openshift/openshift-docs/pull/93895)). Therefore, it should be removed from static entries as well.


SchSeba:
do you need to generate the files again?

PR #150, opened 2025-07-07
Define commatrix files names as consts

shirmoran:
To enable using file names in tests, we rather save the names in consts and reach them that way instead of manually copying the content of the name (as it is now).


aabughosh (commented):
/lgtm
/approve

PR #151, opened 2025-07-08
[release-4.18] Remove bm-sno commatrix from docs

shirmoran:
The bm sno commatrix name has changed to none sno (see [PR](https://github.com/openshift-kni/commatrix/pull/149)), and the pointer to the new commatrix was changed accordingly (see [PR](https://github.com/openshift/openshift-docs/pull/95468#event-18522719853)). Now we can remove the old matrix.


shirmoran:
> are there any ocp docs linking to this that needs to be updated?

They were all updated [here](https://github.com/openshift/openshift-docs/pull/95468).

PR #152, opened 2025-07-14
 OCPBUGS-50488: remove etcd exceptions

tjungblu:
after https://github.com/openshift/cluster-etcd-operator/pull/1444 merges, we can remove those


tjungblu:
@aabughosh the depending PR merged, can we get this in too? thanks

PR #153, opened 2025-07-21
Create a Dockerfile that builds commatrix binary

shirmoran:
This PR creates a Dockerfile that builds the commatrix binary in order to use the commatrix plugin in the origin tests. After the merge of this PR we will add a prow configuration in the release repo in order to build the image and promote it using the ci operator, and than use the build image as a base image for building the origin's repo podman that runs the origin tests (see [dockerfile](https://github.com/openshift/origin/blob/main/images/tests/Dockerfile.rhel)) so the it will contain the necessary plugin.


aabughosh:
look good for me

SchSeba (commented):
If this image is for openshift specific please check the out the images from ART system

for example: https://github.com/openshift/sriov-network-operator/blob/main/Dockerfile.sriov-network-config-daemon.ocp

SchSeba (approved):
/lgtm
/approve

PR #157, opened 2025-08-04
update the image to be used in disconnected env

aabughosh:


pkg/utils/utils.go
```diff
@@ -53,6 +55,40 @@ func (u *utils) WriteFile(path string, data []byte) error {
...
+		nameAndTag = parts[1]
+	case 1:
+		nameAndTag = parts[0]
+	}
+	name, tag, _ := imageutil.SplitImageStreamTag(nameAndTag)
+	return namespace, name, tag
+}
+
+func (u *utils) resolveImageStreamTag(namespace, name, tag string) (string, error) {
+	imageStream, err := u.ImageClient.ImageStreams(namespace).Get(context.TODO(), name, metav1.GetOptions{})
```
SchSeba, line 81: This is not a test please don't use `context.TODO()` you need to use a context with a timeout so the application will not get stuck.

PR #161, opened 2025-08-26
OCPBUGS-59836: Handle services with empty selectors

shirmoran:
Current issue: when the service.Spec.Selector is empty, the labels.SelectorFromSet function return a label that matches all pods. Than, when listing the pods based on the label we check pods which do not match the service.
This issue seems to be the cause of the following bug: [OCPBUGS-59836](https://issues.redhat.com/browse/OCPBUGS-59836)

This PR adds a check whether the service's selector is empty.
if the selector is empty:
- if the service is internal, than skip this service when adding EPS.
- else (extrnal service), we will add all of the ports in the EPS to communication matrix.

if the selector is not empty, we will continue in manner we have checked open ports up until now. 
See the chart explaining the entire flow of the communication matrix creation:
image

### DIFF
--- a/pkg/endpointslices/endpointslices.go
+++ b/pkg/endpointslices/endpointslices.go
@@ -83,29 +83,38 @@ func (ep *EndpointSlicesExporter) LoadExposedEndpointSlicesInfo() error {
 			continue
 		}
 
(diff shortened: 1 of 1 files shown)


SchSeba (approved):
/lgtm
/approve

PR #165, opened 2025-09-03
Remove openshift kube apiserver's static entries

shirmoran:
Ports 6080 and 17697 were added to a service (see [PR](https://github.com/openshift/cluster-kube-apiserver-operator/pull/1875)), so their static entries are no longer needed.


shirmoran:
/cc @SchSeba
Can you please help with the review?

PR #168, opened 2025-09-04
Remove openshift cloud controller manager static entries

shirmoran:
Service was added to port 10258 (see [PR](https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/399)), and the selector of the service of 9258 was updated to match its pod (see [PR](https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/398)). Hence, they can be removed from static entries. 
Also, as there are no more cloud static entries the unit test case which checks the ability to get cloud static entries is not needed any more and was removed.

### DIFF
--- a/cmd/generate/generate_test.go
+++ b/cmd/generate/generate_test.go
@@ -142,8 +142,7 @@ var (
 )
 
 func TestCommatrixGeneration(t *testing.T) {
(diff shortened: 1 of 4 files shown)


SchSeba (approved):
/lgtm
/approve

PR #169, opened 2025-09-04
remove port 10357 as it set as a local port and no need to expose

aabughosh:
remove port 10357 as it set as a local port and no need to expose
bug fix https://github.com/openshift/cluster-kube-controller-manager-operator/pull/848


SchSeba (approved):
/lgtm
/approve

PR #173, opened 2025-09-08
[release-4.20] remove port 10357 as it set as a local port and no need to expose

aabughosh:
remove port 10357 as it set as a local port and no need to expose


SchSeba:
sorry you are right!

/hold cancel

SchSeba:
lets wait to see origin master passing before we merge this one also

PR #181, opened 2025-09-17
adding logs to invistigate the failure of empty noderole

aabughosh:
adding logs to invistigate the failure of empty noderole
https://issues.redhat.com/browse/OCPBUGS-59836
nightlies are failing with that


SchSeba (approved):
/lgtm
/approve

in the future we may want to revert this one if we see it's too chatty

PR #186, opened 2025-09-25
CNF-19535: Generate commatrix by MachineConfig Pools

aabughosh:
We originally generated the commatrix by node role, considering only master and worker nodes. To make it more general, we are now generating it by MachineConfigPool (MCP) to include all roles.

- The code implementation has been updated from NodeToRole to NodeToPool.
- In the static entries, NodeRole is now replaced with the corresponding NodePool name. If a pool contains both master and worker nodes, all relevant static entries for those roles will be included.
- NFT output is now organized by pool.
- The nftables test has been updated to align with applying nftables rules by MCP.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,12 @@ namespace      EndpointSlice Namespace
 service        EndpointSlice owner Service name
 pod            EndpointSlice target Pod name
 container      Port owner Container name
-nodeRole       Service node host role (master/worker/master&worker[for SNO])
+nodePool       Resolved MachineConfigPool name (e.g., master, worker, or custom pool)
 optional       Optional or mandatory flow for OpenShift
 ```
+
+### MachineConfigPool selection
+
+When associating a node to a MachineConfigPool (MCP), the pool is derived directly from the node annotation `machineconfiguration.openshift.io/currentConfig`, expected in the form `rendered-<pool>-<hash>`. The pool name is obtained by removing the `rendered-` prefix and trimming the trailing `-<hash>`.
+
+The resolved pool name is recorded in the `nodePool` field of each matrix entry, and NFT output is generated per pool.
--- a/cmd/README.md
+++ b/cmd/README.md
@@ -57,12 +57,12 @@ Flags:
 Once you run the `oc commatrix generate` command, the plugin will
 generate a communication matrix based on the ingress flows in your
 OpenShift cluster. The output will be saved to a file (destDir) in the chosen format,
-similar to the following:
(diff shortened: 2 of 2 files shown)


cmd/generate/generate_test.go
```diff
@@ -50,6 +51,22 @@ var (
 		},
 	}
 
+	mcpMaster = &machineconfigurationv1.MachineConfigPool{
```
SchSeba: general comment here.

why we are not using ginkgo and gomega like all the other projects?

pkg/mcp/mcp.go
```diff
@@ -0,0 +1,91 @@
...
+			if selector.Matches(labels.Set(node.Labels)) {
+				score := 0
+				if pool.Spec.NodeSelector != nil {
+					score += len(pool.Spec.NodeSelector.MatchLabels)
+					score += len(pool.Spec.NodeSelector.MatchExpressions)
+				}
+				if score > bestScore || (score == bestScore && pool.Name > bestName) {
+					bestName = pool.Name
+					bestScore = score
+				}
```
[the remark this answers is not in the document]
aabughosh: thanks i just implemented on that way

pkg/mcp/mcp.go
```diff
@@ -0,0 +1,118 @@
...
+
+// ResolveNodeToPool builds a mapping from node name to its MachineConfigPool.
+// Selection follows OpenShift MCO precedence rules (https://access.redhat.com/solutions/7004527):
+// 1) If >1 custom pool matches, error
+// 2) If exactly 1 custom pool and master match, error
+// 3) If exactly 1 custom pool matches, select it
+// 4) Else if master matches, select master
+// 5) Else if worker matches, select worker
+// Returns an error if a node does not match any MCP.
+func ResolveNodeToPool(cs *client.ClientSet) (map[string]string, error) {
```
SchSeba: I think there is a much easier way instead of all this API calls and calculation.  you can just take it from the node annotation. example:

```
# oc get node cnfdt4-master-0 -oyaml | grep machineconfiguration.openshift.io/currentConfig

machineconfiguration.openshift.io/currentConfig: rendered-master-054638eb90131424c9e192fda27ec8fa
```

then you just check that split for `-` is 2 and you get the split string in [1] that is the name of the MCP.

aabughosh: @greyerof we can remove the last - and that we will take from 1 to that hash like this return trimmed[:lastDash]

SchSeba (commented):
/lgtm
/approve

/hold

leave to hold until @greyerof has time to review.
feel free to remove the hold after that

PR #191, opened 2025-10-15
Revert "[release-4.20] adding logs to invistigate the failure of empty noderole"

aabughosh:
Reverts openshift-kni/commatrix#183


SchSeba:
please use a cherry pick from the master one

PR #212, opened 2025-10-31
OCPEDGE-2218: feat: add two node arbiter port entries

eggfoobar:
added support for arbiter clusters to the port entries


SchSeba:
@aabughosh please review

PR #275, opened 2025-11-27
Add dynamic ranges to generated communication matrix

shirmoran:
This PR is adding the following changes:

1.  As requested from the partenrs, the dynamic port ranges of the cluster (Linux dynamic\private ranges, kubelet node port dynamic range, host level services dynamic range), are added to the generated communication matrix in all of the supported formats. The dynamic ranges are extracted in the following ways:

- Node port dynamic range: we are looking for a custom defined range (network.Spec.ServiceNodePortRange), if there is no range set, we use a default static range (30000-32767)
- Linux dynamic\private range: we retrive the dynamic range by reading the host sysctl (/proc/sys/net/ipv4/ip_local_port_range).

2. We also have added the option of using a custom entries file that include ranges, so partners will be able to add their own ranges if needed.

3.  The e2e tests have been modified the following:

- EPS vs SS: in the comparsion between the matrices, we also check if host level open ports (from ss matrix) which don't have an EPS are in the range specified in the generated commatrix.
-  Doc vs EPS: in the comparison between the matrices, we also check if the ports in the generated commatrix which are not documented in the doc matrix, do appear in the doc ranges.

4. The custom entries samples had been modified to also include ranges.
5. commatrix.go file's unit tests had been modified to allow mocking of a debugpod in the tests. For that sense, we have added to the commatrixCreator struct a utils field similar to what's done in the ConnectionCheck struct.

### DIFF
--- a/pkg/commatrix-creator/commatrix.go
+++ b/pkg/commatrix-creator/commatrix.go
@@ -9,9 +9,11 @@ import (
 
 	log "github.com/sirupsen/logrus"
 
+	dynamicranges "github.com/openshift-kni/commatrix/pkg/dynamic-ranges"
 	"github.com/openshift-kni/commatrix/pkg/endpointslices"
 	"github.com/openshift-kni/commatrix/pkg/mcp"
 	"github.com/openshift-kni/commatrix/pkg/types"
+	"github.com/openshift-kni/commatrix/pkg/utils"
 	configv1 "github.com/openshift/api/config/v1"
 )
 
@@ -22,9 +24,10 @@ type CommunicationMatrixCreator struct {
 	platformType         configv1.PlatformType
 	controlPlaneTopology configv1.TopologyMode
 	ipv6Enabled          bool
+	utilsHelpers         utils.UtilsInterface
 }
 
-func New(exporter *endpointslices.EndpointSlicesExporter, customEntriesPath string, customEntriesFormat string, platformType configv1.PlatformType, controlPlaneTopology configv1.TopologyMode, ipv6Enabled bool) (*CommunicationMatrixCreator, error) {
+func New(exporter *endpointslices.EndpointSlicesExporter, customEntriesPath string, customEntriesFormat string, platformType configv1.PlatformType, controlPlaneTopology configv1.TopologyMode, ipv6Enabled bool, utilsHelpers utils.UtilsInterface) (*CommunicationMatrixCreator, error) {
 	return &CommunicationMatrixCreator{
 		exporter:             exporter,
 		customEntriesPath:    customEntriesPath,
@@ -32,6 +35,7 @@ func New(exporter *endpointslices.EndpointSlicesExporter, customEntriesPath stri
 		platformType:         platformType,
 		controlPlaneTopology: controlPlaneTopology,
 		ipv6Enabled:          ipv6Enabled,
+		utilsHelpers:         utilsHelpers,
 	}, nil
 }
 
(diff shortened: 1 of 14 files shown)


pkg/commatrix-creator/commatrix.go
```diff
@@ -174,3 +194,150 @@ func expandStaticEntriesByPool(staticEntries []types.ComDetails, poolToRoles map
...
+		return nil, fmt.Errorf("failed to get Linux dynamic private range: %w", err)
+	}
+	dynamicRanges = append(dynamicRanges, linuxDynamicPrivateRange...)
+
+	return dynamicRanges, nil
+}
+
+// GetNodePortDynamicRange returns the cluster's Service NodePort range as dynamic ranges.
+// If the cluster does not define a custom range, it falls back to the Kubernetes default (30000-32767).
+func (cm *CommunicationMatrixCreator) getNodePortDynamicRange() ([]types.DynamicRange, error) {
```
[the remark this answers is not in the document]
shirmoran: Yes okay let's do that. Thanks!

pkg/types/types.go
```diff
@@ -84,6 +95,10 @@ type ContainerInfo struct {
 	} `json:"containers"`
 }
 
+func (dr *DynamicRange) ToString() string {
```
oribon: nit: rename to Ports / PortRange

pkg/types/types.go
```diff
@@ -94,13 +109,34 @@ func (m *ComMatrix) ToCSV() ([]byte, error) {
 		return nil, err
 	}
 
+	// Append dynamic ranges as rows under ComDetails headers.
+	if len(m.DynamicRanges) > 0 {
```
oribon: nit: redundant if

pkg/types/types.go
```diff
@@ -208,6 +249,15 @@ func (m *ComMatrix) ToNFTables() ([]byte, error) {
 		}
 	}
 
+	for _, dr := range m.DynamicRanges {
+		rangeStr := fmt.Sprintf("%d-%d", dr.MinPort, dr.MaxPort)
```
oribon: nit: can we use dr.ToString() here?

pkg/types/types.go
```diff
@@ -339,23 +389,113 @@ func BuildNodeToGroupMap(c rtclient.Client) (map[string]string, error) {
...
+				return nil, fmt.Errorf("invalid port range %q: %w", portStr, err)
+			}
+			ranges = append(ranges, DynamicRange{
+				Direction:   r.Direction,
+				Protocol:    r.Protocol,
+				MinPort:     minPort,
+				MaxPort:     maxPort,
+				Description: r.Service,
+				Optional:    r.Optional,
+			})
```
oribon: can you wrap this in a helper func?

pkg/types/types.go
```diff
@@ -339,23 +389,113 @@ func BuildNodeToGroupMap(c rtclient.Client) (map[string]string, error) {
...
+			Direction: r.Direction,
+			Protocol:  r.Protocol,
+			Port:      port,
+			Namespace: r.Namespace,
+			Service:   r.Service,
+			Pod:       r.Pod,
+			Container: r.Container,
+			NodeGroup: r.NodeGroup,
+			Optional:  r.Optional,
+		})
```
oribon: same (wrap in helper)

pkg/dynamic-ranges/dynamic-ranges.go
```diff
@@ -0,0 +1,167 @@
...
+	minPort, maxPort, err := parsePortRange(rangeStr)
+	if err != nil {
+		log.Errorf("Invalid ServiceNodePortRange format %q: %v", rangeStr, err)
+		return nil, fmt.Errorf("invalid ServiceNodePortRange format %q: %w", rangeStr, err)
+	}
+
+	for i := range dr {
+		dr[i].MinPort = minPort
+		dr[i].MaxPort = maxPort
+	}
```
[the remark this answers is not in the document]
shirmoran: The Idea was to keep all other fields of types.KubeletNodePortDefaultDynamicRange (as dr is set to in line 53) and just change the MinPort and MaxPort field if network.Spec.ServiceNodePortRange is set.
Do you think there is a better way of doing so?

pkg/types/types.go
oribon: when I see a ToString / String method I assume I'm getting a string representation of an object (which could identify it). returning only the ports seems a bit too narrow, considering multiple "DynamicRanges" can return the same

shirmoran: Yeah okay I see your point.
What do you think about "ToPortRangeString"?

cmd/communication-matrix/communication-matrix.csv
oribon: what is this file (and dir) for?

pkg/commatrix-creator/commatrix_test.go
```diff
@@ -289,56 +302,66 @@ var (
...
 	}
+
+	testNetwork = &configv1.Network{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "cluster",
+		},
+		Spec: configv1.NetworkSpec{
+			ServiceNodePortRange: "30000-32767",
+		},
+	}
```
oribon: can you change this to non-default values, and have another test where the field is not specified to see we consume the "default" var?

pkg/dynamic-ranges/dynamic-ranges.go
oribon: can we add some unit tests around the functions here and types.go?

cmd/communication-matrix/communication-matrix.csv
shirmoran: It shouldn't be here. I'll remove it, thanks!

pkg/commatrix-creator/commatrix_test.go
shirmoran: The test doesn't though in the creation of the commatrix we do try to get those values. But we should have desiganted tests for that I agree. I will add them

shirmoran: Though I think we should create a new file for tests under dynamic-ranges, and there we will test the functions  getting the ranges, here we will only mock the behavior the enable the creation of the commatrix (as we do now). What do you think?

oribon (commented):
small comment, and flip the order of 3rd and 2nd commit (so that refactors are first, new thing is last)

PR #400, opened 2026-02-05
Remove Linux Dynamic ranges from generated communication matrix

shirmoran:
This PR removes the linux dynamic ranges from the commatrix, as it includes the entire range of the available ports - hence applying firewall on the matrix with this range will cause the all possible ports open instead of limitting access.

### DIFF
--- a/pkg/commatrix-creator/commatrix_test.go
+++ b/pkg/commatrix-creator/commatrix_test.go
@@ -13,7 +13,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 
 	"github.com/openshift-kni/commatrix/pkg/client"
-	"github.com/openshift-kni/commatrix/pkg/consts"
 	"github.com/openshift-kni/commatrix/pkg/endpointslices"
 	matrixdiff "github.com/openshift-kni/commatrix/pkg/matrix-diff"
 	"github.com/openshift-kni/commatrix/pkg/types"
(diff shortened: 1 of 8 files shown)


SchSeba (commented):
small comment beside that I think we are good

pkg/commatrix-creator/commatrix_test.go
```diff
@@ -652,27 +651,7 @@ var _ = g.Describe("Commatrix creator pkg tests", func() {
...
-					Phase: corev1.PodRunning,
-				},
-			}
-			mockUtils.EXPECT().CreatePodOnNode(gomock.Any(), consts.DefaultDebugNamespace, gomock.Any(), gomock.Any()).Return(mockPod, nil).AnyTimes()
-			mockUtils.EXPECT().DeletePod(mockPod).Return(nil).AnyTimes()
-			mockUtils.EXPECT().WaitForPodStatus(consts.DefaultDebugNamespace, mockPod, corev1.PodRunning).Return(nil).AnyTimes()
-
-			// Mock the command output to return a default port range
-			mockUtils.EXPECT().RunCommandOnPod(mockPod, gomock.Any()).Return([]byte("32768 60999\n"), nil).AnyTimes()
+			// CreateEndpointMatrix lists nodes for MCP-based static entry expansion
```
SchSeba: why this comment is here?

pkg/dynamic-ranges/dynamic-ranges.go
```diff
@@ -6,42 +6,14 @@ import (
...
-	if err != nil {
-		log.Errorf("Failed to get Linux dynamic private range: %v", err)
-		return nil, fmt.Errorf("failed to get Linux dynamic private range: %w", err)
-	}
-	dynamicRanges = append(dynamicRanges, linuxDynamicPrivateRange...)
-
-	return dynamicRanges, nil
-}
-
-// GetNodePortDynamicRange returns the cluster's Service NodePort range as dynamic ranges.
```
SchSeba, line 42: please add this comment to the function about

SchSeba (commented):
just two small nits after that we can merge this

pkg/commatrix-creator/commatrix_test.go
shirmoran: Not needed. Probably forgot to delete it, thanks.

SchSeba (approved):
/lgtm
/approve

PR #424, opened 2026-03-26
Fix nftables test timeout on SNO waiting for MCP update

aabughosh:
WaitForMCPUpdateToStart polled every 30s for a transient state (UpdatedMachineCount != MachineCount). On SNO with NodeDisruptionPolicy, the MCO completes the update in seconds via service reload, so the polling missed the window and timed out after 20 minutes.

Replace WaitForMCPUpdateToStart + WaitForMCPReadyState with a single WaitForMCPUpdate that tracks the rendered MachineConfig name change. This is a persistent state change that cannot be missed regardless of how fast the update completes.

Extract a reusable GetMachineConfigPool helper that returns the full
MachineConfigPool object, and use it in both WaitForMCPUpdate and the
nftables test to avoid duplicated MCP-fetching logic.

### DIFF
--- a/test/pkg/cluster/cluster.go
+++ b/test/pkg/cluster/cluster.go
@@ -27,21 +27,45 @@ const (
 	interval         = 5 * time.Second
 )
 
-func WaitForMCPUpdateToStart(cs *client.ClientSet, role string) {
+// GetMachineConfigPool returns the MachineConfigPool with the given name.
+func GetMachineConfigPool(cs *client.ClientSet, name string) (*machineconfigurationv1.MachineConfigPool, error) {
+	mcp := &machineconfigurationv1.MachineConfigPool{}
+	if err := cs.Get(context.TODO(), controllersClient.ObjectKey{Name: name}, mcp); err != nil {
+		return nil, fmt.Errorf("failed to get %q MachineConfigPool: %w", name, err)
+	}
+	return mcp, nil
+}
+
+// WaitForMCPUpdate waits for the MCO to render a new MachineConfig (by comparing
+// the rendered MC name to previousRenderedMC) and then waits for all machines in the
+// pool to be ready with the new config (timeout: 20m, polling interval: 5s).
+// This avoids polling for transient status changes (UpdatedMachineCount != MachineCount)
+// which can be missed on SNO where NodeDisruptionPolicy completes in seconds.
+func WaitForMCPUpdate(cs *client.ClientSet, name, previousRenderedMC string) {
 	gomega.Eventually(func() (bool, error) {
-		mcp := &machineconfigurationv1.MachineConfigPool{}
-		err := cs.Get(context.TODO(), controllersClient.ObjectKey{Name: role}, mcp)
+		mcp, err := GetMachineConfigPool(cs, name)
 		if err != nil {
-			return false, fmt.Errorf("failed to get %s MachineConfigPool: %v", role, err)
+			return false, err
 		}
 
-		if mcp.Status.UpdatedMachineCount != mcp.Status.MachineCount {
-			log.Printf("MCP %s has started updating", mcp.Name)
+		currentRenderedMC := mcp.Status.Configuration.Name
+		if currentRenderedMC == previousRenderedMC {
+			log.Printf("MCP %s: rendered MC unchanged (%s), waiting for MCO to process", name, currentRenderedMC)
+			return false, nil
+		}
+
+		log.Printf("MCP %s: rendered MC changed from %q to %q", name, previousRenderedMC, currentRenderedMC)
+
+		if mcp.Status.ReadyMachineCount == mcp.Status.MachineCount &&
+			mcp.Status.UpdatedMachineCount == mcp.Status.MachineCount {
+			log.Printf("MCP %s: all machines ready and updated", name)
 			return true, nil
 		}
 
+		log.Printf("MCP %s: still updating (ready=%d, updated=%d, total=%d)",
+			name, mcp.Status.ReadyMachineCount, mcp.Status.UpdatedMachineCount, mcp.Status.MachineCount)
 		return false, nil
-	}, timeout, 30*time.Second).Should(gomega.BeTrue(), "Timed out waiting for MCP to start updating")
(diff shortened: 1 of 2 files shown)


test/pkg/cluster/cluster.go
```diff
@@ -27,21 +27,48 @@ const (
 	interval         = 5 * time.Second
 )
 
-func WaitForMCPUpdateToStart(cs *client.ClientSet, role string) {
+// GetCurrentRenderedMC returns the name of the currently rendered MachineConfig
+// for the given MachineConfigPool role.
+func GetCurrentRenderedMC(cs *client.ClientSet, role string) (string, error) {
```
andreaskaris: nit: role string is the name of the MCP, so better name it 'name string'?
nit: GetCurrentRenderedMCFromPool ? or is that name too long?

andreaskaris: nit: I'm wondering if this function is at all necessary given that it's used only in one location.

Alternatively, I'm wondering if it would make sense to return the MCP, rename the thing to GetMachineConfigPool(...), and then do mcp.Status.Configuration.Name in test/e2e/nftables_test.go and reuse the same function in WaitForMCPUpdate, because there's nearly identical logic there.

So for me, either use no helper function, or use the helper function in both locations.

Something like (typed without testing):
```
func GetMachineConfigPool(cs *client.ClientSet, name string) (string, error) {
	mcp := &machineconfigurationv1.MachineConfigPool{}
	err := cs.Get(context.TODO(), controllersClient.ObjectKey{Name: role}, mcp)
	if err != nil {
		return nil, fmt.Errorf("failed to get %s MachineConfigPool: %v", role, err)
	}
	return mcp, nil
}
```

And:
```
		    mcp, err := cluster.GetMachineConfigPool(cs, pool)
			Expect(err).ToNot(HaveOccurred())
            previousRenderedMC := mcp.Status.Configuration.Name

			machineConfig, err := firewall.CreateMachineConfig(nftablesConfig, artifactsDir,
				pool, utilsHelpers)
			Expect(err).ToNot(HaveOccurred())

			updated, err := cluster.ApplyMachineConfig(machineConfig, cs)
			Expect(err).ToNot(HaveOccurred())

			if updated {
				cluster.WaitForMCPUpdate(cs, pool, previousRenderedMC)
```

And:
```
func WaitForMCPUpdate(cs *client.ClientSet, name, previousRenderedMC string) {
	gomega.Eventually(func() (bool, error) {
		mcp := &machineconfigurationv1.MachineConfigPool{}
		err := cs.Get(context.TODO(), controllersClient.ObjectKey{Name: role}, mcp)
        mcp, err := GetMachineConfigPool(cs, name)
		if err != nil {
			return false, err
		}
		currentRenderedMC := mcp.Status.Configuration.Name
        (...)
```

test/pkg/cluster/cluster.go
```diff
@@ -27,21 +27,48 @@ const (
...
+		return "", fmt.Errorf("failed to get %s MachineConfigPool: %v", role, err)
+	}
+	return mcp.Status.Configuration.Name, nil
+}
+
+// WaitForMCPUpdate waits for the MCO to render a new MachineConfig (by comparing
+// the rendered MC name to previousRenderedMC) and then waits for all machines in the
+// pool to be ready with the new config.
+// This avoids polling for transient status changes (UpdatedMachineCount != MachineCount)
+// which can be missed on SNO where NodeDisruptionPolicy completes in seconds.
```
andreaskaris, line 43: nit: maybe it's useful to document timeout 20 mins and interval 5 secs here, given that they are hardcoded

test/pkg/cluster/cluster.go
```diff
@@ -27,21 +27,48 @@ const (
...
+func WaitForMCPUpdate(cs *client.ClientSet, role, previousRenderedMC string) {
```
andreaskaris: nit: name instead of role?

aabughosh: done

test/pkg/cluster/cluster.go
aabughosh: done

test/pkg/cluster/cluster.go
```diff
@@ -27,21 +27,46 @@ const (
 	interval         = 5 * time.Second
 )
 
-func WaitForMCPUpdateToStart(cs *client.ClientSet, role string) {
+// GetMachineConfigPool returns the MachineConfigPool with the given name.
+func GetMachineConfigPool(cs *client.ClientSet, name string) (*machineconfigurationv1.MachineConfigPool, error) {
+	mcp := &machineconfigurationv1.MachineConfigPool{}
+	err := cs.Get(context.TODO(), controllersClient.ObjectKey{Name: name}, mcp)
```
andreaskaris: nit: sorry, that came from my suggestion. you can combine the if condition here with the assignment
if err := ...; err != nil {}

test/pkg/cluster/cluster.go
```diff
@@ -27,21 +27,46 @@ const (
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get %s MachineConfigPool: %v", name, err)
```
andreaskaris: nit:
```
failed to get MachineConfigPool %s: %w
```

or even
```
failed to get MachineConfigPool %q: %w
```

test/pkg/cluster/cluster.go
```diff
@@ -27,21 +27,46 @@ const (
...
 
-		if mcp.Status.UpdatedMachineCount != mcp.Status.MachineCount {
-			log.Printf("MCP %s has started updating", mcp.Name)
+		currentRenderedMC := mcp.Status.Configuration.Name
+		if currentRenderedMC == previousRenderedMC {
+			log.Printf("MCP %s: rendered MC unchanged (%s), waiting for MCO to process", name, currentRenderedMC)
+			return false, nil
+		}
+
+		log.Printf("MCP %s: rendered MC changed from %s to %s", name, previousRenderedMC, currentRenderedMC)
```
andreaskaris: nit: I prefer quoting with %q but that's absolutely a question of personal preference

PR #427, opened 2026-03-26
Enable unparam check-exported and fix flagged issues

aabughosh:
## Summary
- Enable the `check-exported: true` setting for the `unparam` linter in `.golangci.yml` to catch unused parameters and always-nil error returns in exported functions.
- Fix the three issues it flagged:
  - `commatrix-creator.New()` always returned a nil error — drop the error return value and update all callers.
  - `dynamicranges.GetDynamicRanges()` had an unused `cs` parameter — remove it and update callers.
  - `listeningsockets.NewCheck()` had an unused `destDir` parameter — remove it and update callers.
  - mcp: GetPoolRolesForStaticEntriesExpansion() always returned nil error, drop error return

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -29,6 +29,8 @@ linters-settings:
     disabled-checks:
       - captLocal
       - exitAfterDefer
+  unparam:
+    check-exported: true
   revive:
     rules:
     - name: receiver-naming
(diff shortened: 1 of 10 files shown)


oribon (commented):
/lgtm
/approve

PR #467, opened 2026-04-22
Fix missing space in firewall log prefix for broadcast traffic

### LINKED ISSUE #466: log prefix "firewall" should be "firewall " (with space)
https://github.com/openshift-kni/commatrix/blob/02cf2535ebfb66931844bc9445a94dc486d7a201/pkg/types/types.go#L376

Minor formatting issue:
```
sh-5.1#  journalctl -k --grep firewall
Apr 22 14:53:07 sno10-1 kernel: firewallIN=eno8403np1 OUT= MAC=ff:ff:ff:ff:ff:ff:50:7c:6f:53:d4:1d:08:00 SRC=>
```

andreaskaris:
## Summary
- Add trailing space to the nftables log prefix for broadcast traffic (`ip daddr 255.255.255.255`), matching the existing default drop rule
- Without the space, log entries appear as `firewallIN=...` instead of `firewall IN=...`

Fixes https://github.com/openshift-kni/commatrix/issues/466

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pkg/types/types.go
+++ b/pkg/types/types.go
@@ -373,7 +373,7 @@ table inet openshift_filter {
         udp dport { %s } accept
 
         # Drop broadcast traffic with rate-limited logging
-        ip daddr 255.255.255.255 jump { limit rate 1/minute log prefix "firewall"; drop; }
+        ip daddr 255.255.255.255 jump { limit rate 1/minute log prefix "firewall "; drop; }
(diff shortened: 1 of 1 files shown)


andreaskaris:
@oribon @aabughosh

PR #475, opened 2026-04-27
Skip rpc.statd, dnsmasq and dig in ss vs EPS validation (release-4.21)

aabughosh:
Cherry-pick / equivalent of the validation change from #417 for `release-4.21`.

**Test**
- Skip `rpc.statd` in the EPS matrix diff vs `ss`: ports are nondeterministic when NFS is in use; document expectation is to add those ports or the Linux ephemeral range as extra static entries if needed.
- Skip `dnsmasq` and `dig`: provisioning DHCP/TFTP; not treated as required external listener coverage in this check.

Resolved the failed `git am` on `release-4.21` by applying only these hunks (the upstream patch also touched code that exists only on `main`, e.g. static open-ports `It` and `DynamicRangeList`).

### DIFF
--- a/test/e2e/validation_test.go
+++ b/test/e2e/validation_test.go
@@ -151,6 +151,7 @@ var _ = Describe("Validation", func() {
 			logrus.Warningf("the following ports are not used: \n %s", notUsedEPSMat)
 		}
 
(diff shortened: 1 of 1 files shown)


SchSeba (approved):
/lgtm
/approve 

as a backport (I don't like the idea to pass and also return pointers of the same object)

PR #493, opened 2026-05-15
Fix: expand EndpointSlice entries to all eligible pools based on scheduling constraints

aabughosh:
## Summary

Fixes missing firewall rules when pods reschedule across MachineConfigPools after node reboots.

**Root cause:** EndpointSlice-discovered entries were only assigned to pools where endpoints currently exist (point-in-time snapshot via `getEndpointSliceGroups`). When a workload like the ingress router rescheduled from one worker pool to another after a reboot, the destination pool's nftables rules didn't include the router's ports, causing degraded ClusterOperators (Ingress, OAuth/Authentication, Console).

**Fix:** Instead of looking at where endpoints **are**, we now look at where pods **could be scheduled** by checking their `nodeSelector` and required `nodeAffinity` constraints against all cluster nodes. This ensures firewall rules cover every pool a workload can land on, while still respecting scheduling constraints (constrained workloads only get rules on eligible pools).

## Problem: Before

Consider a cluster with 3 pools: `master`, `worker` (standard), and `customcnf` (custom worker pool). Both `worker` and `customcnf` nodes have the `node-role.kubernetes.io/worker` label.

The ingress router has `nodeSelector: {node-role.kubernetes.io/worker: ""}` and is currently running on the `worker` pool:

```
EndpointSlice endpoints → worker-standard-0 (pool: "worker")
                        
getEndpointSliceGroups() → ["worker"]   ← only current pool

Generated nftables:
  mc-worker.nft:    ports 80, 443, 1936 ✅
  mc-customcnf.nft: (no router ports)   ❌ ← MISSING
```

After a node reboot, the router reschedules to `customcnf` → **ports 80/443/1936 are blocked** → Ingress, OAuth, Console degrade.

## Fix: After

```
Pod nodeSelector: {node-role.kubernetes.io/worker: ""}
Cluster nodes:
  master-0     (pool: "master")  — does NOT match nodeSelector
  worker-std-0 (pool: "worker")  — matches ✅
  worker-cnf-0 (pool: "customcnf") — matches ✅

getEligiblePoolsForPod() → ["customcnf", "worker"]   ← all eligible pools

Generated nftables:
  mc-worker.nft:    ports 80, 443, 1936 ✅
  mc-customcnf.nft: ports 80, 443, 1936 ✅  ← now covered
```

Router can reschedule to any worker pool without breaking firewall rules.

## How scheduling constraints are respected

| Scenario | nodeSelector | Eligible pools |
|----------|-------------|----------------|
| Router (any worker) | `node-role.kubernetes.io/worker: ""` | All worker pools |
| kube-apiserver (masters only) | `node-role.kubernetes.io/master: ""` | Master pool only |
| PTP daemon (specific nodes) | `node-role.kubernetes.io/worker: "", feature.node.kubernetes.io/ptp: "true"` | Only pools with PTP-labeled nodes |
| Unconstrained pod | (none) | All pools |

Tolerations/taints are intentionally not checked: being slightly over-permissive (opening a port on a node where a taint prevents scheduling) is benign, while being under-permissive (missing a firewall rule) causes outages.

## Changes

- **New: `pkg/endpointslices/scheduling.go`** — scheduling constraint matching logic (`getEligiblePoolsForPod`, `nodeMatchesPodConstraints`, nodeAffinity term matching)
- **Modified: `pkg/endpointslices/endpointslices.go`** — store nodes in `EndpointSlicesExporter`; use `getEligiblePoolsForPod` instead of `getEndpointSliceGroups` in `toComDetailsWithGroups`
- **New: `pkg/endpointslices/scheduling_test.go`** — unit tests for all constraint matching scenarios
- **Modified: `pkg/commatrix-creator/commatrix_test.go`** — added multi-pool expansion integration test; made existing test pods more realistic with nodeSelector

## Test plan

- [x] All existing unit tests pass (20/20 in commatrix-creator, 11/11 in endpointslices)
- [x] New scheduling constraint tests cover: no constraints, worker/master nodeSelector, custom labels, required nodeAffinity (In, NotIn, Exists, DoesNotExist), combined constraints, OR-ed selector terms
- [x] New integration test: router on `worker` pool with endpoints only on `standard` → verifies ports appear on both `worker` and `customcnf` pools, but NOT on `master`
- [ ] E2E: generate matrix on cluster with multiple worker MCPs, verify nftables rules include router ports on all worker pools
- [ ] E2E: reboot nodes, verify no ClusterOperator degradation


pkg/endpointslices/endpointslices.go
```diff
@@ -212,8 +213,14 @@ func (ei *EndpointSlicesInfo) toComDetailsWithGroups(nodeToGroup map[string]stri
...
 	}
 
-	// Get the groups backing this EndpointSlice (pool names or roles).
-	pools := ei.getEndpointSliceGroups(nodeToGroup)
+	// Determine all pools where this workload could be scheduled, based on
+	// the pod's nodeSelector and required nodeAffinity. This ensures firewall
+	// rules are created even for pools where no endpoint currently exists but
+	// where a pod could land after rescheduling.
+	pools := getEligiblePoolsForPod(ei.Pods[0], nodes, nodeToGroup)
+	if len(pools) == 0 {
```
andreaskaris: Could you explain, in the comment, the need for this fallback? As it is not clear to me. Thanks!

pkg/endpointslices/endpointslices.go
```diff
@@ -212,8 +213,14 @@ func (ei *EndpointSlicesInfo) toComDetailsWithGroups(nodeToGroup map[string]stri
 		return nil, fmt.Errorf("failed to get pod name for endpointslice %s: %w", ei.EndpointSlice.Name, err)
 	}
 
-	// Get the groups backing this EndpointSlice (pool names or roles).
-	pools := ei.getEndpointSliceGroups(nodeToGroup)
+	// Determine all pools where this workload could be scheduled, based on
+	// the pod's nodeSelector and required nodeAffinity. This ensures firewall
+	// rules are created even for pools where no endpoint currently exists but
+	// where a pod could land after rescheduling.
+	pools := getEligiblePoolsForPod(ei.Pods[0], nodes, nodeToGroup)
```
andreaskaris: I forgot how endpointslices work, exactly, but is there ever a case where an endpointslice could match pods with different YAML definitions, and thus ei.Pods[0] could have a specific node selector, but ei.Pods[1] could have a different node selector? Because I think that's possible?

pkg/endpointslices/scheduling.go
```diff
@@ -0,0 +1,99 @@
...
+	slices.Sort(pools)
+	return pools
+}
+
+// nodeMatchesPodConstraints checks whether a node satisfies the pod's
+// nodeSelector and required nodeAffinity. Tolerations/taints are intentionally
+// not checked: being slightly over-permissive (opening a port on a node where
+// a taint would prevent scheduling) is benign, while being under-permissive
+// (missing a firewall rule) causes outages.
+func nodeMatchesPodConstraints(pod corev1.Pod, node *corev1.Node) bool {
```
andreaskaris: I think this entire function's content can be replaced with:
```
	nodeaffinity "k8s.io/component-helpers/scheduling/corev1/nodeaffinity"
(...)
// nodeMatchesPodConstraints checks whether a node satisfies the pod's
// nodeSelector and required nodeAffinity. Tolerations/taints are intentionally
// not checked: being slightly over-permissive (opening a port on a node where
// a taint would prevent scheduling) is benign, while being under-permissive
// (missing a firewall rule) causes outages.
// If no NodeSelector or RequiredDuringScheduling Affinity is specified, the pod
// matches all nodes.
func nodeMatchesPodConstraints(pod corev1.Pod, node *corev1.Node) bool {
	// Check if a pod's node affinity + nodeSelector matches a node:
	req := nodeaffinity.GetRequiredNodeAffinity(&pod)
	match, err := req.Match(node)
	if err != nil {
		return false
	}
	return match
}
```

As `GetRequiredNodeAffinity`  checks both nodeSelector and nodeAffinity and it returns `true` when both are nil (see the req.Match) source coe
```
// GetRequiredNodeAffinity returns the parsing result of pod's nodeSelector and nodeAffinity.
```

I mean, at that point, you can also get rid of that helper altogether and just implement it in `getEligiblePoolsForPod`. Which IMO would just look like:
```
	nodeaffinity "k8s.io/component-helpers/scheduling/corev1/nodeaffinity"
(...)
func getEligiblePoolsForPod(pod corev1.Pod, nodes []corev1.Node, nodeToGroup map[string]string) []string {
	poolsMap := make(map[string]struct{})
	req := nodeaffinity.GetRequiredNodeAffinity(&pod)
	for i := range nodes {
		if match, err := req.Match(&nodes[i]); err != nil || !match {
			continue
		}
		if pool := nodeToGroup[nodes[i].Name]; pool != "" {
			poolsMap[pool] = struct{}{}
		}
	}

	pools := slices.Collect(maps.Keys(poolsMap))
	slices.Sort(pools)
	return pools
}
```

And make sure that the tests for `getEligiblePoolsForPod` match all corner cases such as no selectors / affinity, or preferredaffinity set, set.

andreaskaris: Perhaps also add a comment about PreferredDuring.. - given that these pods can still run anywhere, we don't have to worry about that and they are the same as pods without selectors to us.

pkg/endpointslices/scheduling_test.go
```diff
@@ -0,0 +1,360 @@
+package endpointslices
+
+import (
+	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestGetEligiblePoolsForPod_NoConstraints(t *testing.T) {
```
andreaskaris:
```
Please rewrite these as TestGetEligiblePoolsForPod(t *testing.T) {
  // use test cases instead of multiple functions
}
```

pkg/endpointslices/scheduling_test.go
```diff
@@ -0,0 +1,360 @@
...
+		},
+	}
+
+	pools := getEligiblePoolsForPod(pod, nodes, nodeToGroup)
+	if len(pools) != 1 || pools[0] != "standard" {
+		t.Fatalf("expected [standard], got %v", pools)
+	}
+}
+
+func TestNodeMatchesPodConstraints_BothNodeSelectorAndAffinity(t *testing.T) {
```
andreaskaris:
```
Please rewrite these as TestNodeMatchesPodConstraints(t *testing.T) {
  // use test cases instead of multiple functions
}
```

Also add a test for the case that the nodeselector and required affinity are not set (= matching all nodes), and another one with preferred affinity set (that shows that preferred affinity has not impact)

pkg/endpointslices/scheduling.go
```diff
@@ -0,0 +1,99 @@
...
+		if !nodeMatchesPodConstraints(pod, &nodes[i]) {
+			continue
+		}
+		pool := nodeToGroup[nodes[i].Name]
+		if pool != "" {
+			poolsMap[pool] = true
+		}
+	}
+
+	pools := make([]string, 0, len(poolsMap))
```
andreaskaris: I think this is more idiomatic in new versions of go, I certainly find it easier to read:
```
	pools := slices.Collect(maps.Keys(poolsMap))
```

pkg/endpointslices/scheduling.go
```diff
@@ -0,0 +1,99 @@
...
+
+// getEligiblePoolsForPod returns all pools where the pod could potentially be
+// scheduled, based on its nodeSelector and required nodeAffinity constraints.
+// Unlike getEndpointSliceGroups which only returns pools where endpoints
+// currently exist, this function returns all pools that have at least one
+// node matching the pod's scheduling constraints. This ensures that firewall
+// rules are generated for every pool a workload could land on after
+// rescheduling (e.g., after a node reboot).
+func getEligiblePoolsForPod(pod corev1.Pod, nodes []corev1.Node, nodeToGroup map[string]string) []string {
+	poolsMap := make(map[string]bool)
```
andreaskaris: nit: I'd use an empty struct here `map[string]struct{}`, just because it takes up no space and you're not using the bool in an if/else for convenience; it's personal preference, though. The struct's advantage is that it takes no space, however :)

pkg/endpointslices/scheduling.go
```diff
@@ -0,0 +1,99 @@
...
+	for i := range nodes {
+		if !nodeMatchesPodConstraints(pod, &nodes[i]) {
+			continue
+		}
+		pool := nodeToGroup[nodes[i].Name]
+		if pool != "" {
```
andreaskaris: if pool := nodeToGroup[nodes[i].Name]; pool != "" {

pkg/endpointslices/endpointslices.go
```diff
@@ -212,8 +213,21 @@ func (ei *EndpointSlicesInfo) toComDetailsWithGroups(nodeToGroup map[string]stri
...
+	// For scheduler-managed pods, determine all pools where the workload could
+	// be scheduled based on nodeSelector and required nodeAffinity, so that
+	// firewall rules are created for pools where a pod could land after
+	// rescheduling (e.g., after a node reboot).
+	var pools []string
+	if isStaticPod(ei.Pods[0]) {
+		pools = ei.getEndpointSliceGroups(nodeToGroup)
+	} else {
+		pools = getEligiblePoolsForPod(ei.Pods[0], nodes, nodeToGroup)
+		if len(pools) == 0 {
```
andreaskaris: Pleae explain in the comments why this if condition here is needed, as I don't understand it

pkg/endpointslices/endpointslices.go
```diff
@@ -212,8 +213,21 @@ func (ei *EndpointSlicesInfo) toComDetailsWithGroups(nodeToGroup map[string]stri
...
+	// be rescheduled, so the EndpointSlice snapshot is authoritative for them.
+	// For scheduler-managed pods, determine all pools where the workload could
+	// be scheduled based on nodeSelector and required nodeAffinity, so that
+	// firewall rules are created for pools where a pod could land after
+	// rescheduling (e.g., after a node reboot).
+	var pools []string
+	if isStaticPod(ei.Pods[0]) {
+		pools = ei.getEndpointSliceGroups(nodeToGroup)
+	} else {
+		pools = getEligiblePoolsForPod(ei.Pods[0], nodes, nodeToGroup)
```
andreaskaris: what happens if ei.Pods[0] and ei.Pods[1] have different selectors? I know a service can have Pods with different definitions and nodeselectors. Can this happen for your endpointslices? would it make sense to iterate over all pods inside getEligiblePoolsForPod(s)?

pkg/endpointslices/endpointslices.go
```diff
@@ -212,8 +213,21 @@ func (ei *EndpointSlicesInfo) toComDetailsWithGroups(nodeToGroup map[string]stri
...
-	// Get the groups backing this EndpointSlice (pool names or roles).
-	pools := ei.getEndpointSliceGroups(nodeToGroup)
+	// Static pods (owner Kind == "Node") are managed by the kubelet and cannot
+	// be rescheduled, so the EndpointSlice snapshot is authoritative for them.
+	// For scheduler-managed pods, determine all pools where the workload could
+	// be scheduled based on nodeSelector and required nodeAffinity, so that
+	// firewall rules are created for pools where a pod could land after
+	// rescheduling (e.g., after a node reboot).
+	var pools []string
+	if isStaticPod(ei.Pods[0]) {
```
andreaskaris: Instead of the nested if:
```
var pools []string
if !isStaticPod(ei.Pods[0]) {
    pools = getEligiblePoolsForPod(ei.Pods[0], nodes, nodeToGroup)
}
if len(pools) == 0 {
    pools = ei.getEndpointSliceGroups(nodeToGroup)
}
```

pkg/endpointslices/endpointslices.go
andreaskaris: then, obviously, in the rest of the function (and for isStaticPod specifically), we're using  the first pod only, as well

andreaskaris (changes requested):

pkg/endpointslices/scheduling_test.go
andreaskaris: actually not needed if you drop that helper altogether, then just integrate the tests in the one for GetEligible

pkg/endpointslices/endpointslices.go
aabughosh: this is needed because pools can be empty in two cases:

Static pods -- hasStaticPod skips getEligiblePoolsForPods so pools stays nil. For static pods the EndpointSlice snapshot is authoritative since they can't be rescheduled.
Edge case -- getEligiblePoolsForPods ran but no node matched the pod's constraints. We fall back to the EndpointSlice data so we don't generate zero firewall rules.

pkg/endpointslices/endpointslices.go
aabughosh: Static pods -- hasStaticPod returns true, so getEligiblePoolsForPods is skipped and pools stays nil. For static pods (kubelet-managed, can't be rescheduled), the EndpointSlice snapshot is the correct source of truth.
Edge case -- getEligiblePoolsForPods ran but no node matched the pod's constraints (e.g. node labels changed). Falling back to the EndpointSlice data ensures we still generate rules for pools where traffic is currently flowing

pkg/endpointslices/endpointslices.go
aabughosh: done

pkg/endpointslices/scheduling.go
aabughosh: thanks much better

pkg/endpointslices/scheduling.go
aabughosh: done

pkg/endpointslices/endpointslices.go
```diff
@@ -212,8 +213,28 @@ func (ei *EndpointSlicesInfo) toComDetailsWithGroups(nodeToGroup map[string]stri
...
+	// 1. Static pods (skipped above): they are pinned to a specific node by
+	//    the kubelet, so the point-in-time EndpointSlice snapshot is the
+	//    correct source of truth.
+	// 2. Scheduler-managed pods whose constraints didn't match any node
+	//    (e.g. the node labels changed, or custom groups aren't reflected
+	//    in nodeToGroup). Using the EndpointSlice data as a safe fallback
+	//    ensures we still generate rules for pools where traffic is flowing.
+	if len(pools) == 0 {
+		pools = ei.getEndpointSliceGroups(nodeToGroup)
+	}
```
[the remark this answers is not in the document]
andreaskaris: this is actually what I was worried about, too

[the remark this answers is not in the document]
andreaskaris: perhaps use a function filterStaticPods or something and then act on each group (static pods and non-static). this corner case (static and non-static pods matched by the same svc) is also so unlikely, though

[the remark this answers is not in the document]
aabughosh: done

aabughosh:
@coderabbitai

pkg/endpointslices/scheduling.go
```diff
@@ -0,0 +1,64 @@
...
+// Tolerations/taints are also not checked: being slightly over-permissive
+// (opening a port on a node where a taint would prevent scheduling) is
+// benign, while being under-permissive (missing a firewall rule) causes
+// outages.
+func getEligiblePoolsForPods(pods []corev1.Pod, nodes []corev1.Node, nodeToGroup map[string]string) []string {
+	poolsMap := make(map[string]struct{})
+	for _, pod := range pods {
+		req := nodeaffinity.GetRequiredNodeAffinity(&pod)
+		for i := range nodes {
+			if match, err := req.Match(&nodes[i]); err != nil || !match {
```
andreaskaris, line 53: nit: I actually think we can do `for _, node := range nodes`  (sorry)
but no need to change this, either is fine

pkg/endpointslices/endpointslices.go
aabughosh: done

pkg/endpointslices/scheduling_test.go
```diff
@@ -0,0 +1,332 @@
...
+						"node-role.kubernetes.io/worker": "",
+						"feature.node.kubernetes.io/ptp": "true",
+					},
+				},
+			}},
+			nodes: []corev1.Node{
+				{ObjectMeta: metav1.ObjectMeta{Name: "master-0", Labels: map[string]string{"node-role.kubernetes.io/master": ""}}},
+				{ObjectMeta: metav1.ObjectMeta{Name: "worker-0", Labels: map[string]string{
+					"node-role.kubernetes.io/worker": "",
+					"feature.node.kubernetes.io/ptp": "true",
```
andreaskaris: I think you can add this label to the defaultNodes and all other tests should still work, and you can use defaultNodes here, too

andreaskaris (commented):
2 nits, feel free to ignore.

I'll review it again a bit later today, I need to do a second pass

but overall I think it looks good, I just want to understand it well

Edit:  I'll review it again tomorrow

pkg/commatrix-creator/commatrix_test.go
```diff
@@ -994,4 +996,194 @@ var _ = g.Describe("Commatrix creator pkg tests", func() {
...
+			masterNode := &corev1.Node{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "master-0",
+					Annotations: map[string]string{
+						"machineconfiguration.openshift.io/currentConfig": "rendered-master-abc",
+					},
+					Labels: map[string]string{"node-role.kubernetes.io/master": ""},
+				},
+			}
+			standardWorker := &corev1.Node{
```
andreaskaris, line 1018: qq: wouldn't, technically, the standardWorker and the customcnfWorker, both be matched by the same MCPs (standard and customcnf)? both have selectors for `MatchLabels: map[string]string{"node-role.kubernetes.io/worker": ""}`  (nothing that makes this test invalid, it's just a bit less realistic)

pkg/endpointslices/endpointslices.go
```diff
@@ -212,8 +213,24 @@ func (ei *EndpointSlicesInfo) toComDetailsWithGroups(nodeToGroup map[string]stri
...
+	// could be scheduled based on nodeSelector and required nodeAffinity,
+	// so that firewall rules cover every pool a pod could land on after
+	// rescheduling (e.g. after a node reboot).
+	nonStaticPods, hasStatic := filterNonStaticPods(ei.Pods)
+	pools := getEligiblePoolsForPods(nonStaticPods, nodes, nodeToGroup)
+
+	// Include the EndpointSlice-observed groups when static pods are present
+	// (the snapshot is their source of truth) or as a safe fallback when the
+	// constraint-based expansion returned nothing (e.g. node labels changed).
+	if hasStatic || len(pools) == 0 {
```
andreaskaris: Actually ... Is this `if` condition needed / beneficial, or could it be detrimental, sometimes?

My thinking:
Scenario a) You get the eligible pools for all nonstaticpods. If a single static pod is present, then you also merge in the getEndpointSliceGroups, and you remove the duplicates.
Scenario b) You get the eligible pools for all nonstaticpods. But you don't find any. So you get the endpointslicegroups instead as a fallback and use those. So clearly, we believe that there are scenarios where the eligible pools for nonstaticpods do not capture all needed pools.

Are we sure that getEligiblePoolsForPods is always a superset of getEndpointSliceGroups (the fact that we have a fallback makes me believe the contrary)? Or are there cases where the 2 might diverge, yes len(pools) > 0?
What if you simply:
Do getEligiblePoolsForPods(nonStaticPods) and then merge this with getEndpointSliceGroups() always, without the if condition?

Just thinking out loud.

andreaskaris (commented):
/lgtm but please see my nits and especially my comment wrt the if condition `if hasStatic || len(pools) == 0 {` as I'm wondering if it's needed and if we shouldn't simply merge all the time, without the if

pkg/endpointslices/endpointslices.go
aabughosh: you are right

pkg/endpointslices/scheduling_test.go
aabughosh: done

pkg/commatrix-creator/commatrix_test.go
aabughosh: done

pkg/endpointslices/endpointslices.go
```diff
@@ -212,8 +213,22 @@ func (ei *EndpointSlicesInfo) toComDetailsWithGroups(nodeToGroup map[string]stri
...
 
-	// Get the groups backing this EndpointSlice (pool names or roles).
-	pools := ei.getEndpointSliceGroups(nodeToGroup)
+	// For scheduler-managed pods, determine all pools where the workload
+	// could be scheduled based on nodeSelector and required nodeAffinity,
+	// so that firewall rules cover every pool a pod could land on after
+	// rescheduling (e.g. after a node reboot). Static pods (owner Kind ==
+	// "Node") are filtered out because they are pinned to a node by the
+	// kubelet and cannot be rescheduled.
+	nonStaticPods, _ := filterNonStaticPods(ei.Pods)
```
andreaskaris: nit: let's get rid of the boolean (second argument) in filterNonStaticPods and I think then we're good

aabughosh: done

andreaskaris (approved):
/lgtm

